PR #9, opened 2015-11-01
Show local IP address in STA mode

zaporylie:
Would be nice to print IP address to console.

BTW Nice library - thanks ;)


tzapu:
nice, thanks

PR #20, opened 2015-12-14
Allocate the DNSServer and ESP8266WebServer on the heap at begin() time and free them when autoConnect() terminates.

alltheblinkythings:
This reduces the likelihood of a stack overflow when WifiManager is instantiated in setup(),
and lowers memory usage (after WiFi is connected) when it's instantiated globally.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -13,13 +13,7 @@
 
 #include "WiFiManager.h"
 
-
-//DNSServer dnsServer;
-
-// Web server
-//ESP8266WebServer server(80);
-
-WiFiManager::WiFiManager() : server(80) {
+WiFiManager::WiFiManager() {
 }
 
 void WiFiManager::begin() {
@@ -27,6 +21,8 @@ void WiFiManager::begin() {
 }
 
 void WiFiManager::begin(char const *apName) {
+  dnsServer.reset(new DNSServer());
+  server.reset(new ESP8266WebServer(80));
   
   DEBUG_PRINT("");
   _apName = apName;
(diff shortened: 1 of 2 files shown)


tzapu:
wo, thanks very much for this. you obviously know a looooot more about how all this works than me so i had some questions:
- would i need to include memory.h in the example/main sketches as well?
- do these 
```
server.reset();
dnsServer.reset();
```
clear all the allocated memory ?
- i assume it lowers memory consumption because the servers are not allocated until it can't actually connect to wifi, rather than on initialisation, is that correct?

thank you for this, much appreciated

alltheblinkythings:
* You shouldn't need to `#include <memory>` anywhere.  It's included in WiFiManager.h, so everything else that includes that will automatically understand the std::unique_ptr<x> syntax.  (I just tested and confirmed the examples compile.)

* `std::unique_ptr<X> my_var;` is a C++ templated way of saying "`delete` an object of type X" when `my_var` goes out of scope" (in this case, "scope" is as long as WiFiManager exists, because ), and `my_var.reset();` is a way of saying `delete X` now, even though WiFiManager still exists.  Provided that the author of type X (in this case, a `ESP8266WebServer` or `DNSServer`) did the right thing in their destructor (e.g. [`ESP8266WebServer::~ESP8266WebServer`](https://github.com/esp8266/Arduino/blob/a5c3aea98ee4cb445b973d2970dcbeefc52216c1/libraries/ESP8266WebServer/src/ESP8266WebServer.cpp#L64)), it will free all memory used by those objects.  (It actually looks like DNSServer might not do this correctly, I'll send a patch for that to the Arduino tree).

* There are two changes to memory consumption.  Yes, it lowers memory consumption when not used because it saves constructing the servers until the call to begin(...) when the initial wifi connection fails.  And it frees memory once autoConnect() has returned, even if someone wrote code like this with a global WiFiManager, meaning that the WiFiManager would never be destroyed:
 ```
#include <WiFiManager.h>

WiFiManager wm;

setup() {
  wm.autoConnect();
}
 ```
The second change that comes from using std::unique_ptr is that that the memory allocated to store the `ESP8266WebServer` and `DNSServer` is allocated on the heap (which is roughly "all the device's free memory"), rather than on stack (which is about 4kB of RAM, I think).  In my case, the extra memory used by the two servers made me run out of stack and my device would crash whenever I tried to add a WiFiManager in setup(), hence this change.

tzapu:
wow man, simply amazing, thank you very much.
I will test it later and merge it. 
thank you again, this is great

alltheblinkythings:
No problem.  And thanks for writing this in the first place!!!  It made the job of setting up something and handing it to a friend for them to configure on their network ridiculously easy!  :+1:

PR #28, opened 2015-12-22
Add multiple nets

domonetic:


tzapu:
this might cause a bit of wear and tear on the flash, wouldn't the wifi main library write the ssid and pass at every attempt to connect?
thank you very much for this, i will think on it a bit more as there is also this 
https://github.com/esp8266/Arduino/tree/master/libraries/ESP8266WiFi/examples/WiFiMulti
which deals with multiple known networks in a different way

domonetic:
If you're right, maybe find another way not saved to flash

lunanigra:
Thanks so far! I tried following update...
https://github.com/domonetic/WiFiManager/tree/132b223f7d7331c0a1305c93620095a0bc0b1de3

Perhaps I missed something; but I get compiler error. First WifiManager.h seems to need some modifications as well. Also the definition of the method connectWifi needs an update as no return value is returned (but required).

domonetic:
Sorry. I coding without test.

PR #40, opened 2015-12-30
add library.properties

domonetic:
to standardize the library and appear in Library Manager


domonetic:
ouchhh, yes I thought was a more extensive list

domonetic:
changed

PR #44, opened 2016-01-04
Manually starting the configuration portal

riccibitti:
My hardware has a button to trigger configuration, so I needed a way to force the captive portal regardless of running configuration. So I added a third (defaulted) parameter to autoconnect(), namely forceAP. It is defaulted to false, for backward compatibility; but if it set to true, the captive portal is started without even attempting to connect with current SSID and password.

In the works, I have cleaned up some commented/unused code and changed connectWifi() to return the connection status, getting rid of temprary 's' varibles that were used to read status again right after a call to connectWifi().

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -16,14 +16,6 @@
 WiFiManager::WiFiManager() {
 }
 
-void WiFiManager::begin() {
-  begin("NoNetESP");
-}
-
-void WiFiManager::begin(char const *apName) {
-  begin(apName,NULL);
-}
-
 void WiFiManager::begin(char const *apName, char const *apPasswd) {
   dnsServer.reset(new DNSServer());
   server.reset(new ESP8266WebServer(80));
@@ -113,7 +105,6 @@ boolean WiFiManager::autoConnect(char const *apName, char const *apPasswd) {
   connect = false;
   begin(apName,apPasswd);
 
-  bool looping = true;
   while(timeout == 0 || millis() < start + timeout) {
     //DNS
     dnsServer->processNextRequest();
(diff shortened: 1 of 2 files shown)


tzapu:
very nice, thanks
do you mind if I edit it a bit, i'd like to rename the parameter to skipConnect, or something like that, for more 'legibility'

riccibitti:
No problem for editing or renaming, although personally I would not chose "skipConnect" as new name because it is implementation-related (the Arduino IDE does not even expose libraries internal code), and also autoconnect(skipConnect) sounds weird to me. Not to mention that the function exits establishing a connection anyway!
I would choose something more focused on the results the user expects, like "configureAlways" or the like.
Maybe a more elegant solution would be abstracting the captive portal thing from the autoconnect function, that way the captive portal configuration can be invoked from autoconnect() (in case of failed connection) but also externally from user code e.g. triggering a button.
End of academic-mode talking! 
Congratulations for a very nice and useful piece of code.

tzapu:
you of course are correct. 
I was thinking changing autoConnect to something like begin or connect, but to be honest, the more i think about it, the less i can make a case for it. can t think why begin would be better than autoConnect.

thank you very much, i ll add this soon, need to sort out some other changes first

domonetic:
all library in arduino call begin, only for compatibility i think!!

riccibitti:
Sorry for my english, I like the `autoconnect` name, I agree `begin` would not work as well in my usage scenario, that is, invoking the WiFiManager portal when a button is pressed.

Please let the code speak for me:

```
void setup() {
  pinMode( BUTTONPIN, INPUT_PULLUP );
  WiFi.mode(WIFI_STA);
  WiFi.begin();
  ..other initializations here..
}

void loop() {
  if ( digitalRead(BUTTONPIN)==LOW )
  {
    WiFiManager wifiManager;
    wifiManager.autoConnect( NULL, NULL, true );
    WiFi.mode(WIFI_STA);
    WiFi.begin();
  }
  
  ..other housekeeping here..
}
```
My point was about `skipConnect` as parameter name in place of e.g.`configureAlways`, I mean, if I think at the parameters when invoking the function, what I see is:

```
    wifiManager.autoConnect(  
          /*apName=*/NULL, 
          /*apPasswd=*/NULL, 
          /*configureAlways=*/true );
```
for me, the name above it is easier to understand than:
```
    wifiManager.autoConnect(  
          /*apName=*/NULL, 
          /*apPasswd=*/NULL, 
          /*skipConnect=*/true );
```

Hope this clarifies the point. 
By the way, I have the feeling that the additional argument can be made unnecessary factoring the code as described in my previous post, hopefully during the weekend I will have enough time to attempt it and make the code speak clearer than me ;-)

tzapu:
@riccibitti you are right, i just now understood what you meant in the previous post
will update it to start the portal in another function, and then you can call that directly from the button calling code

tzapu:
merged changes and created a new function
`startConfigPortal(apName, apPassword = NULL);`
that you should be able to call to setup config mode on demand

give it a try if you can

tzapu:
added an example as well
there are some issues with the esp sdk currently ,and you need to do some weird stuff like add WiFi.mode(WIFI_STA); before everything, otherwise you can t connect.
the example is functional as it is
cheers

riccibitti:
Great, I'll test it right away

PR #48, opened 2016-01-06
allow custom parameters

tzapu:
> I saw some comments on the plans to add custom parameters.

> Here I'm contributing a working implementation, it uses the EEPROM.

> Keep the nice work on this library!

> Just my 2 cents.

i hope you don't mind, i created a new pull request on a secondary branch so i can play with it a bit without disrupting the main branch.

absolutely love what you ve done, thank you

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -13,7 +13,95 @@
 
 #include "WiFiManager.h"
 
+WiFiManagerParameter::WiFiManagerParameter(const char *id, const char *placeholder, const char *defaultValue, int length){
+	_id = id;
+	_placeholder = placeholder;
+	_length = length;
+	_value = new char[length];
+	for(int i=0; i<length; i++)
+		_value[i] = 0;
+	if(defaultValue!=NULL)
+		strncpy(_value, defaultValue, length);
+}
+
+const char* WiFiManagerParameter::getValue(){
+	return _value;
+}
+const char* WiFiManagerParameter::getID(){
+	return _id;
+}
+const char* WiFiManagerParameter::getPlaceholder(){
+	return _placeholder;
+}
+int WiFiManagerParameter::getValueLength(){
+	return _length;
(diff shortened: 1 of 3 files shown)


domonetic:
this is an optional? I use EEPROM for other data and do not want me remaining place.

tzapu:
yup, optional, don't want to add more dependencies, i am trying to modify it to abstract all saving to the app instead of the library

domonetic:
deberia tener como parametro la posicion de inicio de la eeprom

tzapu:
i have modified the pull request so the storage is left to the user. i don t want to re add the eeprom dependency to the lib. especially since people using eeprom already will have their own mechanism of saving data.
if you can see and test the version here
https://github.com/tzapu/WiFiManager/tree/dev
it has a config.json FS storage example. will make an eeprom example as some point soon as well.
will leave it a couple of days in case anyone has any feedback, then will cleanup and merge.

thank you very much @rpseng and @domonetic , much appreciated

domonetic:
Sorry but I do not see sense. 
Why would it profit wifimanager store more parameters that are not using in its purpose that is connected to wifi?
Using the library json do that or a simple struct can handle extra information without making the library grow

tzapu:
the sole benefit is that it adds the input fields in the configuration page, so you use the same input page/mechanism for everything at once.
If you need it it s very useful (i'd like to save my emoncms/blynk tokens in the page rather than hardcoding). 
If you don't need it, what it adds currently is insignificant, it doesn't include any extra libraries, it's just a few lines of code in the main lib. Hence my option of not having eeprom or fs included in the lib itself.
i hope i managed to be more clear now.

tzapu:
"Using the library json do that or a simple struct can handle extra information without making the library grow" 
you can do that with the new changes anyway, saving and storing is not implemented in WiFiManager anymore, just the display and data input.

PR #52, opened 2016-01-09
Adding a method to get the config portal SSID

franklinvv:
I wanted to use the config portal SSID in the callback, so I can direct a user to that SSID. I added getConfigPortalSSID() to enable that.


tzapu:
thank you, 
added an example as well, worth rethinking this, as as it is, it forces you to allocate WiFiManager as a global, which wastes memory

franklinvv:
You're absolutely right. I'll fix this right away.

tzapu:
you sir are a wonderful human being :D

PR #53, opened 2016-01-09
On-demand example now compatible with ESP-01 boards

riccibitti:
Changed to make it work on ESP-01 boards.
It turn out that on ESP-01, the pin cannot be checked at setup as its only available pins, gpio0 and gpio2, are shared with the bootloader and **must** be kept HIGH at power-up.
- moved triggering of configuration portal into Arduino's loop() function
- changed pin from 4 to 0 (and made it  #define-able)

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -49,13 +49,11 @@ void WiFiManager::addParameter(WiFiManagerParameter *p) {
   DEBUG_WM(p->getID());
 }
 
-void WiFiManager::setupConfigPortal(char const *apName, char const *apPassword) {
+void WiFiManager::setupConfigPortal() {
   dnsServer.reset(new DNSServer());
   server.reset(new ESP8266WebServer(80));
 
   DEBUG_WM(F(""));
-  _apName = apName;
-  _apPassword = apPassword;
   start = millis();
 
   DEBUG_WM(F("Configuring access point... "));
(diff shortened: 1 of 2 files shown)


riccibitti:
I'm new to github and messed quite a bit with the merges, hope everything is ok! 
In the end the only file intended to change is **OnDemandConfigPortal.ino**

tzapu:
@riccibitti i ve put it on 4 on purpose, you wouldn't be able to use pin 0 or 2 on an ESP01 like that because as far as i remember, if either one of them is low at boot, the esp module won't boot in the proper mode anymore (either programming mode, or sd card mode)
for the esp01 you'd need a whole different example, one that checks the pin later than in the boot phase...

riccibitti:
@tzapu  Please note that OnDemand *is working* on my ESP-01 ! 
You are correct, with  this push I moved the check from `setup()` to `loop()`.
I just forgot to mention that I placed a *pushbutton*, in place of a jumper/switch. I connected it to pin 0,  but I expect pin 2 to work as well (will test it tonight).

ESP-01 is immensely popular, I think it's important to offer ESP-01 users the capability to use OnDemand, maybe splitting the example in a "switch" ( `setup()` ) variety and a "button" ( `loop()`, ESP-01 compatible) flavour?

tzapu:
you are correct, did not notice you move it to loop

i wonder if i can merge your commits in only one somehow... :P

tzapu:
could you please rebase so all your changes are 1 commit with rebase?
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

tzapu:
i think i can do it... lemme try :d

tzapu:
and i couldn't clean it up :p
merged, thank you very much

riccibitti:
Finally got to rebase, guess too late... Tanks anyway for the pointer to the docs.
As for the Arduino, I'm using 1.6.6 on Windows 10 with board toolchain from http://arduino.esp8266.com/stable/package_esp8266com_index.json

tzapu:
no probs, you did a such a great job of this, thanks
1.6.6 i think introduced a new arduino_builder script that is smarter and can included the needed modules. It also seems to have a lot of other issues so people still use 1.6.5 which as far as i remember needs the headers included both in lib and main sketch. 
i ve re aded them just to be safe

PR #55, opened 2016-01-09
Added parameter to callback

franklinvv:
To show the config portal SSID in the callback, a global reference to the WiFiManager object was needed. To solve this, a pointer to the WiFiManager object is now passed to the callback. The feedback example is edited accordingly.

This does break backward compatibility though, but saves memory.


tzapu:
perfect :D
thank you

PR #64, opened 2016-01-17
Adding Static IP Support

jonathanendersby:
As a  side note, I think we should give _ip etc more explicit names to avoid confusion.


tzapu:
yes, that would be good

i will test and add as soon as i can, really good job.

i m thinking i should expose the ip settings through the web portal as well at some point.

i still can't believe how awesome this community is and how much support and changes i've received from people.

THANKS

PR #74, opened 2016-01-19
@PlatformIO Library Registry manifest file

ivankravets:
* This library in Web Registry: http://platformio.org/#!/lib/show/567/WifiManager
* Specification: [library.json](http://docs.platformio.org/en/latest/librarymanager/config.html)
* Library Manager:  http://docs.platformio.org/en/latest/librarymanager/index.html

### DIFF
--- a/WifiManager
+++ b/WifiManager
@@ -1,0 +1,13 @@
+{
+  "name": "WifiManager",
+  "keywords": "wifi, wi-fi",
+  "description": "ESP8266 WiFi Connection manager with fallback web configuration portal",
+  "repository":
+  {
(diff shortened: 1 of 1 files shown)


doom369:
@ivankravets I just was thinking to write to you about it =).

ivankravets:
@doom369 I've just received request from user as for this lib :)

P.S: Do you use @PlatformIO for building/uploading for @blynkkk?

doom369:
Not at the moment. @PlatformIO in my todo list.

ivankravets:
@doom369 Thanks! :+1:

PR #75, opened 2016-01-19
Specify valid compatible platform ID and remove unsupported "architecture" field

ivankravets:
`library.json` specification http://docs.platformio.org/en/latest/librarymanager/config.html


ivankravets:
Reflected in https://github.com/tzapu/WiFiManager/pull/76

PR #108, opened 2016-02-19
Added an option to the WiFiManagerParameter so that a form parameter can be…

PabluskiNC:
Added an option to the WiFiManagerParameter so that a form parameter can be read-only. This is done in a way that does not break an existing program by having a constructor for either way. 
If a "1" is added to the end of the parameter list then the field becomes readonly.
The HTML will add the "readonly" option.

My need for this is that I am using the MAC address as a ClientID in a MQTT application. I wanted the 
user to know the mac without being able to modify it.

It is a relatively simple change, but I hope that you can use it.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -12,7 +12,8 @@
 
 #include "WiFiManager.h"
 
-WiFiManagerParameter::WiFiManagerParameter(const char *id, const char *placeholder, const char *defaultValue, int length) {
+  
+WiFiManagerParameter::WiFiManagerParameter(const char *id, const char *placeholder, const char *defaultValue, int length, int rofield) {
   _id = id;
   _placeholder = placeholder;
   _length = length;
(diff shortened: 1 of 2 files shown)


tzapu:
hi, sorry for missing this
did you check the info page? i think i added the mac address there

pieman64:
@tzapu I couldn't see any reference to MAC addresses on your info page.

tzapu:
that's weird, they are in the code
```
  page += F("<dt>Soft AP MAC</dt><dd>");
  page += WiFi.softAPmacAddress();
  page += F("</dd>");
  page += F("<dt>Station MAC</dt><dd>");
  page += WiFi.macAddress();
  page += F("</dd>");
```

are they not appearing?

pieman64:
Sorry I thought you were referring to your Github information page rather than the AP page.

PabluskiNC:
The pull request is not so much about the MAC address as it is about being able to have a readonly parameter.

tzapu:
hi, just thinking if it s not better to add then some new type of parameter that would cause basically a label (or html snippet) to show up and be ignored in the posting of the page
since this is for display pourposes alone, it might be faster, better and more functional to add it like that.
people then can add legends anwhatever else they desire...

PabluskiNC:
That sounds interesting. Maybe a generic function, sunch as htmlLine, that basically passes whatever text you want as HTMl and maybe has a few "tags" that can be replaced with certain parameters (such as IP, mac, etc.).

tzapu:
thanks for this, modified it a bit and you can use it like this:
```

  WiFiManagerParameter custom_mqtt_server("server", "mqtt server", "iot.eclipse", 40, " readonly");
  WiFiManagerParameter custom_mqtt_port("port", "mqtt port", NULL, 5);
  WiFiManagerParameter custom_text("<p>This is jsut a text paragraph</p>");
  WiFiManagerParameter custom_blynk_token("blynk", "blynk token", "sdfsfasdfdsfsdfa", 32);

  wifiManager.addParameter(&custom_mqtt_server);
  wifiManager.addParameter(&custom_mqtt_port);
  wifiManager.addParameter(&custom_text);
  wifiManager.addParameter(&custom_blynk_token);
```

PR #112, opened 2016-02-26
Jump to form for password entry

tablatronix:
If you have a long list of APs, the password field gets focused on click, but this happens below the fold, this commit adds a jump to the form so you can see it. Added id to form instead of jumping to the #p field because it makes the input too close to the top of page, a well known issue of anchor links.


tablatronix:
I don't know why this can't auto merge and has conflicts, i just forked and edited...

tablatronix:
I am gonna delete this, I think i reforked and github didnt warn me and kept old parent.

tzapu:
that s a cool fix, if you want i can add it directly so you don t jump through the hoops again

tablatronix:
I added #113 to replace this.
I do not usually use the web interface for this work, but thought i would try and github lets you fork something you already forked so i just assumed my fork was up to date, and it was not, it was not merged since i last forked it, odd bug there. But its amazing what you can do now with just the web interface.

PR #113, opened 2016-02-26
Jump to form for password entry

tablatronix:
If you have a long list of APs, the password field gets focuses on click, but this happens below the fold, this jumps to the form so you can see it. Added id to form instead of jumping to the #p field because it makes the input too close to the top of page, a well known issue of anchor links.


tablatronix:
I might have to look at this, in chrome, it still focuses the password field and i can type BUT i get no cursor in the input now.

hmmm

tablatronix:
Change to jump to the #p password field, keeps the focus. alternative would be way to much work.

PR #118, opened 2016-03-06
give option for static OR dynamic

richardwilliamson:
adds an option to set enableForceStaticIP which when true will always
present the static IP fields - if no static IP is sent these will be
blank. If the user doesn’t fill these in then dhcp will be used. Also
added “getSTAIsStaticIP()” - call this after the save is completed and
query whether the user has set a static IP. This way the sketch can
then save the static IP to EEPROM if required.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -215,6 +215,7 @@ int WiFiManager::connectWifi(String ssid, String pass) {
   DEBUG_WM(F("Connecting as wifi client..."));
 
   // check if we've got static_ip settings, if we do, use those.
+  //TODO - should we be checking that we have a gw and sn? What happens if not?
   if (_sta_static_ip) {
     DEBUG_WM(F("Custom STA IP/GW/Subnet"));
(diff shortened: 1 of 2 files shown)


tzapu:
hi, except the dynamic switch between static and dynamic (which i did not look, but i think if you empty the fields, it will go dynamic anywya) all the features to do this are already in the lib
i don t see the benefit, yet, of adding yet another way of doing it

richardwilliamson:
Hello, thanks for your response. My problem with the current lib is that the IP fields only seem to be displayed if a static IP is sent to the library - meaning that the user has no way to switch between static and DHCP. With my branch you can force the display of these two fields, and then query whether the user has set a static IP so the code can then save the static IP to EEPROM if required. 

Thanks!

baruch:
This is indeed needed. The user should be able to dynamically control if he wants a static IP and then the fields need to be displayed and used.

tzapu:
ok guys, point made, will be included, just got a bit more cleanup and releases to do before i can go through it

thank you again very much for everything

PR #119, opened 2016-03-06
Fix non saving

richardwilliamson:
Added a WiFi.disconnect() if we are using DHCP which seems to fix the
problem with wifi not connecting when using a manually called portal.
Also added an option to force a save (and exit) even if the wifi
doesn’t connect - this is useful if using on a device away from the
network that it will be eventually used on

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -178,7 +178,7 @@ boolean  WiFiManager::startConfigPortal(char const *apName, char const *apPasswo
       DEBUG_WM(F("Connecting to new AP"));
 
       // using user-provided  _ssid, _pass in place of system-stored ssid and pass
-      if (connectWifi(_ssid, _pass) != WL_CONNECTED) {
+      if (connectWifi(_ssid, _pass) != WL_CONNECTED  && !_forceSaveOnDone) {
         DEBUG_WM(F("Failed to connect."));
       } else {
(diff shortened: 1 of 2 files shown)


tzapu:
i have added some changes that might fix some connecting issues in the latest couple of commits
if you could try them in your scenario it would be much appreciated

this bit
```
 if (_sta_static_ip) {
    DEBUG_WM(F("Custom STA IP/GW/Subnet"));
    WiFi.config(_sta_static_ip, _sta_static_gw, _sta_static_sn);
    DEBUG_WM(WiFi.localIP());
  } else
  {
	DEBUG_WM(F("DHCP - disconnect"));
	WiFi.disconnect();
  }
```
will make connecting the first time impossible if not using a static ip as disconnect deletes ssid/pass and this won t work
```
 DEBUG_WM("Using last saved values, should be faster");
 WiFi.begin();
```

richardwilliamson:
Have added a few more features now..

- There is a method to constantly run the wifiManager in the background, meaning that the user can config at any time
- There is a way to upload the sketch through wifiManager
- Added a refresh (with 5 sec delay) to the save page, maybe this should link to a page which displays a save status (succeeded or failed) - either way it's quite useful
- Added a callback so that you can update parameters on page load (due to running in background)

PR #124, opened 2016-03-09
Remove duplicate aps

### LINKED ISSUE #26: Duplicate Wifi networks
Hello, when scanning in our company all Wifi networks are listed 2-3 times.

I didn't had the chance for further investigation; but I assume following reason... We have several access points providing the same Wifi network. Means we can use network XYZ inside the whole building on all floors. So, assumption is that a Wifi network is getting listed multiple times if multiple access points are in range.

Is there a chance putting all items with the same name together?

Thanks, JC

tablatronix:
fixes #26


tablatronix:
ready now

tzapu:
awesome, thanks
will pull it in a bit later

PR #128, opened 2016-03-20
Adds option to use generated AP name but with password

javl:
This update allows you to use a generated AP name, but also set a password.
`wifiManager.ssidFromId()` returns the generated AP name, so you can use it in the original `autoConnect()` functions:

    // use generated AP name, no password (as before)
    wifiManager.autoConnect();
    // Use a generated AP name, but also a password
    wifiManager.autoConnect(wifiManager.ssidFromId(), "AP-PASSWORD");

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,11 +97,14 @@ if you just want an unsecured access point
 ```cpp
 wifiManager.autoConnect("AP-NAME");
 ```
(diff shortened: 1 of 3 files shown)


javl:
Closing this pull request as I need to test it first; resulted in a weird AP name

PR #129, opened 2016-03-20
Small changes

javl:
- Added toggle to credential form to show / hide password when entering it.
- Credential form now used POST instead of GET
- Adds pointer cursor when hovering over button
- Adds .gitignore to hide .DS_Store files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
--- a/WiFiManager.h
+++ b/WiFiManager.h
@@ -25,17 +25,17 @@ extern "C" {
 }
 
 const char HTTP_200[] PROGMEM             = "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\n\r\n";
-const char HTTP_HEAD[] PROGMEM            = "<!DOCTYPE html><html lang=\"en\"><head><meta name=\"viewport\" content=\"width=device-width, initial-scale=1, user-scalable=no\"/><title>{v}</title>";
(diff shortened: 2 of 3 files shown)


tzapu:
i ll also need to make post/get a setting somwhere, some people like get , some like post :)

javl:
Not sure how using GET would be helpful. Best approach to setting the credentials through a direct URL would be to create an endpoint specifically for this purpose I think.
    192.168.4.1/set/<ssid>/<password>

tzapu:
GET, as it currently works, sends them as params

tzapu:
also, is there any particular reason on moving all these buttons to post as well?
```
"<form action='/wifi' method='POST'><button>Configure WiFi</button></form><br/><form action='/0wifi' method='POST'><button>Configure WiFi (No Scan)</button></form><br/><form action='/i' method='POST'><button>Info</button></form><br/><form action='/r' method='POST'><button>Reset</button></form>";
```

tzapu:
thanks for the pull request and changes, btw

javl:
I just changed them all to post, don't think it matters but might be cleaner to at least be consistent.

javl:
To quickly explain:
My thinking was that using POST is cleaner, is it does not leave the credentials in your browser history as a human readable URL, the way GET does.
I can imagine it would be useful to enter a URL including the params to set the credentials directly, but you might be better of having this as an optional way. So the default way is to use the form, with POST, and the alternative way could be to call a page with the GET parameter included;

    192.168.4.1/setManual?s=<ssid>&p=<password>

tzapu:
well, some people asked me specifically not to change to POST because they like having the vars visible, probably for debug etc)
i guess everyone has their preference.
i will move to POST using your pull request, thanks, but i ll add a switch for toggle between post and get.

i will also revert the pages to GET, as post adds unneeded complications (for instance if you want to go back, or refresh the wifi page, it will ask you if you want to submit form again)

is that ok?

baruch:
I would think that for an embedded system it would be better to choose one way and stick to it. Reduces code size and complexity. I personally do not mind between GET and POST.

javl:
I agree. Maybe it would be useful to ask WHY they want Get? I can't really think of a good reason, but they might have a specific use in mind.

tzapu:
well, the expressed preference was that they could specifically see the password in the history, double check it was what it was supposed to, etc...

i think it can work just as well with get and post, the get param functions i think are agnostic of protocol, so there isn t much complexity really...

igrowing:
My 2 cents for POST beside "human readable" address mentioned by javl:
Browser keeps all the details in address history. Once you submitted wrong details (mistake happens) browser will propose this link on next address input. If you mistaken 2-3 times the errors go behind visible area in the URL box. So end-user has all chances to go around his mistakes like a dog for its tail.
IMHO, the software should be user friendly as possible.

By code size/complexity, it doesn't matter for MCU what type of message it receives.

tablatronix:
I also agree data should not be submitted via gets, post options default is better.

kentaylor:
@javl would you be willing to add this to the https://github.com/kentaylor/WiFiManager/ branch?

tablatronix:
I agree, change to POST, add api endpoints for querystring api control

tablatronix:
It looks like the webserver uses REQUEST so you can still send it info however you want. This just affects the form.

petslane:
@tzapu 
> [...]
> i will also revert the pages to GET, as post adds unneeded complications (for instance if you want to go back, or refresh the wifi page, it will ask you if you want to submit form again)

To avoid submitting form again, server must respond to POST request with 3xx redirect response to `/wifi` or `/0wifi` page.

PR #145, opened 2016-04-01
added a callback function for the information page.

flok99:
this allows a sketch to do some basic checking if all works as expected so that you do not need the hassle of first configuring wifi, finding the ip-address, and so on. maybe there are other uses as well. Please note: I did not test if it works, this commit is only to give an idea of what I would like.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -60,7 +60,7 @@ const char* WiFiManagerParameter::getCustomHTML() {
   return _customHTML;
 }
 
-WiFiManager::WiFiManager() {
+WiFiManager::WiFiManager() : ipcb(NULL) {
 }
 
 void WiFiManager::addParameter(WiFiManagerParameter *p) {
@@ -621,6 +621,13 @@ void WiFiManager::handleInfo() {
   page += F("<dt>Station MAC</dt><dd>");
   page += WiFi.macAddress();
   page += F("</dd>");
+  if (ipcb) {
+     char buffer[128];
+     ipcb(buffer, sizeof buffer);
+     page += F("<dt>Extra info:</dt><dd>");
+     page += String(buffer);
+     page += F("</dd>");
+  }
   page += F("</dl>");
   page += FPSTR(HTTP_END);
 
--- a/WiFiManager.h
+++ b/WiFiManager.h
@@ -24,6 +24,8 @@ extern "C" {
   #include "user_interface.h"
 }
 
+typedef void (* InfoPageCallback)(char *, size_t);
+
 const char HTTP_200[] PROGMEM             = "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\n\r\n";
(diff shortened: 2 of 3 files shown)


tzapu:
hi, why not just run all checks before starting up WiFiManager? outside of it pretty much

flok99:
It is not about checks. It is about being able to tell the user of the
device e.g. which device it is or maybe some example measurements so that
he can see if the device is ok.


On Tue, May 24, 2016 at 7:35 AM, tzapu <notifications@github.com> wrote:

> hi, why not just run all checks before starting up WiFiManager? outside of
> it pretty much
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/tzapu/WiFiManager/pull/145#issuecomment-221170925>
>



-- 
www.vanheusden.com  www.slimwinnen.nl

tzapu:
hi, you can run checks, do what you need, and display them in the connect page ...
would that not achieve the same thing... you can show pretty much whatever you want as custom html, checks, information, etc

tablatronix:
I suppose you are referring to

`WiFiManagerParameter custom_text("<p>This is just a text paragraph</p>");
wifiManager.addParameter(&custom_text);`

But maybe a callback could be required for data that you cannot generate before hand ?
Something that uses data from wifi or the connection, shrug.

tzapu:
in that particular case, maybe some kind of generic purpose button in menu/generated page might be better 

like you did, but more generic oriented, so people coould add their own pages to the menu…

worth a thought

sbeland2:
I am trying to add an input field to the page (email address field), but then when I do that, the wifi connection to the ap cannot connect, if I don't fill it, then it connect?!

here how i did it:

='password'><input id='e' name='e' length=128 placeholder='email address'><br/>";

I want to be able to grab an email address from the user so i can send him an email to say that I connected successfully (in fact, it will connect to my server and create a user device entry), from there, he will be able to setup whatever.

I tried also to get what is saved to eeprom, or at least get it after it connect, unsuccessfull, here my code:

void WiFiManager::ListParams (void)
{
  //parameters
    if (_params[i] == NULL) {
      break;
    }
    //read parameter
    String value = server->arg(_params[i]->getID()).c_str();
    //store it in array
    value.toCharArray(_params[i]->_value, _params[i]->_length);
    DEBUG_WM(F("Parameter"));
    DEBUG_WM(_params[i]->getID());
    DEBUG_WM(value);
  }			

and i call this one from the setup once it connect successfully.

sbeland2:
i noticed that in esp8266wifi.h the ssid and password are set to 16 char!!!
what happens if an ap is using 32 char?
the max size of ssid (based on IEEE standard) is 32, so would need char ssid [33];

  char _ssid[16];
    char _password[16];
    
    bool startLocalAp();
    bool startLocalServer();
    char _localAPSSID[16];
    char _localAPPassword[16];

kentaylor:
In the comment https://github.com/tzapu/WiFiManager/pull/145#issuecomment-244670891 @sbeland2  seems to be refering to a different repository, perhaps this one - https://github.com/ekstrand/ESP8266wifi/blob/master/README.md

tablatronix:
#452

tablatronix:
- [ ] change PR to development branch
- [ ] remove the static html, implement only the callout, user adds their own content

PR #157, opened 2016-04-24
Major Functionality Change To Improve Reliability

kentaylor:
The WiFiManager library has been unreliable for me.

I think this is due to the incorrect premise that the ESP8266 should be instructed to connect to a WiFi network. This has led to the library including unneeded functionality which has the side effect of causing the ESP8266 module to intermittently fail to connect to WiFi, behave inconsistently and occasionally brick. 

If this pull request is accepted the library functionality will be substantially changed and it would be included into sketches in a different way.


tzapu:
oh man, this is such a massive amount of effort into work and documenting stuff i ve put into this... thank you, very much appreciated.

i am sorry for not replying sooner, also it might be a while until i ll manage to look at it properly.

i am simply way to tied up currently with stuff and as the issues board might show, i have simply not chance of offering this the attention it deserves.

I would love to discuss about the way it works, the issues the current model has, the motivation between doing it as it is, etc and hopefully i ll get out of this problem period sooner rather than later

sorry again for the delays

kentaylor:
The reply was soon enough, thanks. I've some sympathy for the effort required. I was expecting to have network connectivity sorted in an afternoon and it has now consumed many hours over weeks. Particularly as it was failing in strange unrepeatable ways. I liked your battery powered temperature sensor too as I have a similar [goal.] (http://blog.urremote.com/2016/01/how-hot-is-lounge.html)

As it is a major change and because the claimed failure modes are intermittent and tricky to test I've provided some supporting evidence at  https://github.com/kentaylor/WiFiManager#evidence-the-premise-is-wrong-and-the-problems-it-causes .

It would be great to get some more people trialing the proposed changes as my arguments need to be tested. Then when you get more time you can have confidence in whether never calling WiFi.begin() when it is already trying to auto connect in the background can improve reliability.

tablatronix:
I will take a look when i can

WiFiManager.cpp
```diff
@@ -278,40 +335,37 @@ uint8_t WiFiManager::waitForConnectResult() {
...
-  }
-  return _ssid;
-  }
+  DEBUG_WM(F("END WPS"));
+}
+//Convenient for debugging but wasteful of program space.
+//Remove if short of space
+char* WiFiManager::getStatus(int status)
+{
+   switch (status)
```
tzapu: it seems that the statuses returned by ESP8266WIFISTA are a bit different from what i can tell
```
/**
 * Return Connection status.
 * @return one of the value defined in wl_status_t
 *
 */
wl_status_t ESP8266WiFiSTAClass::status() {
    station_status_t status = wifi_station_get_connect_status();

    switch(status) {
        case STATION_GOT_IP:
            return WL_CONNECTED;
        case STATION_NO_AP_FOUND:
            return WL_NO_SSID_AVAIL;
        case STATION_CONNECT_FAIL:
        case STATION_WRONG_PASSWORD:
            return WL_CONNECT_FAILED;
        case STATION_IDLE:
            return WL_IDLE_STATUS;
        default:
            return WL_DISCONNECTED;
    }
}
```

tzapu:
hello, i finally managed to take a better look at this, and will look some more, but i have some more questions and clarifications:
-this was meant to be as simple as possible, take out most of the complexity on setting up WiFi for the user, this being the reason for the blocking flow and autoConnect and a lot of other decisions
-as far as i can see, this mostly does what https://github.com/tzapu/WiFiManager/blob/master/examples/OnDemandConfigPortal/OnDemandConfigPortal.ino was doing, with, of course, a lot of extras and refinements
-most of my sensors do not have a button
-some of those sensors are running on battery, adding a 60 second delay on every wake up, would be a killer of battery life
-a lot of the issues as you say, is mostly calling WiFi.begin(), why change the whole flow, remove autoConnect(), etc ? 
-i d like to mention this again, my goal is to keep this as lightweight and as simple as possible, somebody that picks it up for the first time, together with picking up arduino ide for the first time, should be able to use it in 30 seconds. power users will be able to modify and suit their needs anyway.

thanks for doing this, looking forward to your reply
cheers

WiFiManager.cpp
```diff
@@ -136,24 +168,44 @@ boolean WiFiManager::autoConnect(char const *apName, char const *apPassword) {
...
-  WiFi.mode(WIFI_AP_STA);
-  DEBUG_WM("SET AP STA");
-
+  int connRes = WiFi.waitForConnectResult();
+  if (connRes == WL_CONNECTED){
+	  WiFi.mode(WIFI_AP_STA); //Dual mode works fine if it is connected to WiFi
+	  DEBUG_WM("SET AP STA");
+  	}
+  	else {
+    WiFi.mode(WIFI_AP); // Dual mode becomes flaky if not connected to a WiFi network.
```
tzapu: it seems this is mostly because the ESP8266 can only use one channel, so if the AP i started on channel 1 and the network you are connecting to is on channel 5, it will drop the connections from the AP as it switches all to channel 5
there might be other issues as well, but this would explain a lof of my observations on this

WiFiManager.cpp
```diff
@@ -201,50 +255,47 @@ boolean  WiFiManager::startConfigPortal(char const *apName, char const *apPasswo
...
   return  WiFi.status() == WL_CONNECTED;
 }
 
-
 int WiFiManager::connectWifi(String ssid, String pass) {
-  DEBUG_WM(F("Connecting as wifi client..."));
-
+  DEBUG_WM(F("Connecting wifi with new parameters..."));
+  if (ssid != "") {
+	resetSettings(); /*Disconnect and wipe out old values. If you don't do this
```
tzapu: once you do this, if i haven t completly lost the plot, leaving the ssid and password field empty in the config portal, like instructed in the screenshot, will never result in reusing the credentials

examples/ConfigOnDoubleReset/ConfigOnDoubleReset.ino
```diff
@@ -0,0 +1,104 @@
+/*
```
tzapu: this is an absolutely brilliant idea. 
thank you

WiFiManager.cpp
kentaylor: All of the values returned by ESP8266WIFISTA are correctly identified. However there is some extras which are WL_NO_SHIELD, WL_SCAN_COMPLETED, WL_CONNECTION_LOST. I'll remove them in the next version and if any extra values are added to ESP8266WiFiSTAClass::status() they will become the default value which is UNKNOWN

WiFiManager.cpp
kentaylor: resetSettings is only called when a new SSID value has been provided because of the line - if (ssid != "")

Therefore previous settings remain unchanged if new ones are not provided.

WiFiManager.cpp
kentaylor: Yes, apart from the issue you highlighted, switching channels causes confusion elsewhere. When scanning for WiFi networks the ESP drops the WiFi network to scan the other channels. The browsing device sometimes connects to a different WiFi network automatically before the AP WiFi is restarted. Users don't notice the network change and it confuses them. This is more likely when there is a high density of WiFi networks. The AP WiFi network also disappears when it needs to change channels to connect to a WiFi network and reappears on the new channel. Browsing devices will often switch WiFi networks when this occurs, again unnoticed and causing confusion.

The scanning issue can be improved by scanning before setting up the WiFi network and then listing networks from memory rather than re-scanning. 

It is also why an app is easier to use than a browser because, unlike a browser, an app can control  which network it connects to and switch back and forth as required.

examples/ConfigOnDoubleReset/ConfigOnDoubleReset.ino
kentaylor: Yes, a nice contribution from @datacute. I updated the documentation to reflect this change. Looking at some of the other @tzapu   comments.

> -most of my sensors do not have a button

I can't think of any other way to get human initiation which I think is necessary and the @datacute solution is good if the sensors have a reset button. 

> -some of those sensors are running on battery, adding a 60 second delay on every wake up, would be a killer of battery life

Yes I don't like that one either. I see someone else made a non blocking configuration portal but I'm not keen on that because it changes timing within the user program which is often problematic for IOT applications. 

> -a lot of the issues as you say, is mostly calling WiFi.begin(), why change the whole flow, remove autoConnect(), etc ? 

I set out the reasons for [deprecating  autoConnect()](https://github.com/kentaylor/WiFiManager#configuration-portal-initiation-not-automatic) . I find that it works well during development but when deployed it is rare that new credentials are required when there is a loss of WiFi network visibility.

> -id like to mention this again, my goal is to keep this as lightweight and as simple as possible, somebody that picks it up for the first time, together with picking up arduino ide for the first time, should be able to use it in 30 seconds. power users will be able to modify and suit their needs anyway.

Yes. The [autoConnect example](https://github.com/tzapu/WiFiManager/blob/master/examples/AutoConnect/AutoConnect.ino) is wonderfully simple. I wonder if autoconnect could be redefined to mean start up a configuration portal if no SSID has been defined. That would avoid getting a configuration whenever the WiFi network fails and retains the simplicity.

WiFiManager.cpp
kentaylor: Modified scanning to take place before configuration portal is created for browser interface, JSON interface unchanged. 

This avoids the client device becoming disconnected from the WiFi network during scanning. 

There was also an issue previously that, because scanning took a long time, users would refresh, go back, click on other selections etc in the belief nothing was happening. This would build up a backlog of requests on the web server that it would queue and carry out sequentially but for which the client device was no longer waiting for a response, thus causing confusion.

examples/ConfigOnDoubleReset/ConfigOnDoubleReset.ino
tablatronix: I have devices like this , to reset or force config we have to press reset within the first 5 seconds ( a status light blinks ) and do so 5 times in a row, it is a pretty decent method to prevent accidental glitch wipes etc. ( this is commercial equipment )

tablatronix:
These examples are way to advanced for , if this is to be merged I suggest adding back the original ones, and adding these long verbose ones into a new example folder called advanced or something.

tablatronix:
Not bad seems to work well.

@tzapu 
I have some suggestions and cleanups for it if you merge I can fix it up a bit, merge it into a new branch at least for starters so the other pull requests can be built on it.

I think it is still pretty lightweight, most of the bloat can be ignored.

* minor changes to template and verbage and UI
 * Exit Portal is actually Close or Stop Portal, shutdown etc
 * Saving flow could be more concise, simply confirm saved ( does it confirm save, or just say it saved ?Then continue, or auto meta redirect ?Device is now connected to AP SSID with IP of 10.0.0.0 etc.
* minor document changes, cleanup of course, reformat , remove references to "new" functionality and redocument as fact.
* remove the app references, keep json, but remove references to the app and dns name feature, not needed. Mention in readme as footnote, untested unsupported etc.
* wrap debugging in compile definitions, so they are not in compiled code unless defined WMDEBUG enabled, else use debug constant during runtime, either works, this might already exist.

tablatronix:
` I wonder if autoconnect could be redefined to mean start up a configuration portal if no SSID has been defined. `

Good idea, would fix the need for wrapping it with connected or SSID check, and allow ship to config easier.

tzapu:
thank you @tablatronix and @kentaylor 
i have started porting this together with a cleanup in the development branch, but as expected, goes slower than i would have wanted due to time constraints
they are all very good points and want to get to them all, but i don t want to drop the autoConnect behaviour or examples as they are as a lot of people use it like that, including myself. 
docs and examples could explain all a bit more structured, as a lot stuff (start portal only if SSID is not defined, on button press config, on double reset, etc) are already possible in normal version of WIFIManger with just a bit of extra code
for example, for doing `I wonder if autoconnect could be redefined to mean start up a configuration portal if no SSID has been defined` , all that is needed would be similar to:
```
If (WiFi.SSID != "") { 
  wifiManager.autoConnect()
}
```
again thank you for all your points, work and observations, they are really inspiring

examples/ConfigOnDoubleReset/ConfigOnDoubleReset.ino
kentaylor: Yes @tablatronix , but I'd be surprised if the commercial device resets though. 

A led that flashes faster while the button is pressed, eventually becoming solid is what the [Google Chromecast](https://www.google.com/intl/en_au/chromecast/tv/chromecast/) does. I like it and copied the idea in [this sketch](https://github.com/kentaylor/esp8266-weather-station/blob/master/examples/WeatherDHT22ThreeScreensWiFiManager/WeatherDHT22ThreeScreensWiFiManager.ino) where a progress bar on a screen is used to launch the configuration portal. In the sketch it uses this [button library](https://github.com/kentaylor/OneButton) that can detect press, double press, long press and long press followed by still pressed. The button becomes multi purpose then.

I was thinking of adding it with flashing led instead of progress bar as a WiFiManager example that behaves like a Chromecast but it's a bit of effort.

kentaylor:
@tablatronix  said:-

> Saving flow could be more concise, simply confirm saved ( does it confirm save, or just say it saved ?Then continue, or auto meta redirect ?Device is now connected to AP SSID with IP of 10.0.0.0 etc.

I'd agree that this is the weakest area of the interface but doing better requires more than

> minor changes to template

@tablatronix doesn't like the app reference but the app is the only way to handle this aspect of the interaction well. The response page is returned before flash is updated which it must, because the web server and WiFi network shuts down when it tries to connect to the AP. The data returned must be bland because we haven't tried to connect to an AP yet and there are lots of reasons it might fail e.g. wrong password. The WiFi can be down for a long time, upwards of 30 seconds under some circumstances and the browsing device will usually auto switch to another network when the WiFi is down too long and/or the ESP changes WiFi channels. An IPhone will close the config portal if that occurs which minimises confusion but browsing devices all do things differently.
. 
Unlike an app there is no way to control a WiFi connection from a browser. The idea at the moment is to have a URL which is valid whichever network the browsing device is on so that it doesn't look like the ESP device has failed when the browsing device fails to connect to the device URL. Once you have experience with something you can not see it like a naive user but users interpreting a failed URL as device failure was what I usually saw in user testing. This functionality could also be implemented with JavaScript. JavaScript can't query WiFi network status but could probe URL's to see if it got a response from the internet or the device and then display an appropriate message.

@tablatronix  said:-

> How about also adding a erase flash command also

This is already there and documented on the information page where all valid commands are shown.

kentaylor:
@tzapu the idea of autoconnect is that when an ESP device starts up a user can configure a WiFi network if the ESP can't connect to a WiFi network . This is fabulously simple to include in a sketch but I don't like it because in practice I've found more often that not the absence of the already configured WiFi network is temporary and a configuration portal isn't useful. I've taken the view that it isn't possible for the device to know the difference but an algorithm that would self correct in the case of a temporarily absent network would be to keep checking for WiFi connectivity once the configuration portal has been launched and close it automatically if WiFi becomes connected.

tablatronix:
Yes I agree with the above also, I see wifi manager as a one time or problem or changing networks only thing, not a launch when connection fails always. I doubt many people will actually us it as such.

I want the app removed because it creates a dependancy for this library and an expectation, and that is bad for oss, tapzu and contributers should not have to worry about breaking third party dependancies, but I think the json api is nice to have and good enough for users to attempt to use if they chose to. I just do not think a static url and app should be tied to it at all. Unless there is some standard local mdns way to do so , there is not that i know of.

tablatronix:
@tzapu I see 2 branches, neither has any cherry picking or merges from this that I can see, do you have milestones or an issue created for these branches so we can follow or contribute, are you planning on manually implementing these and not merging ? Trying to understand what the 2 branches are and what you propose. Its still your repo and baby.

oh and this lib was just mentioned on adafruits show and tell, so you might see more popularity.

tablatronix:
@kentaylor 
>This is already there and documented on the information page where all valid commands are shown.

I didn't see anything about that in your readme. How can you add a reset/clear button to the main page ?

umranium:
@tablatronix @tzapu @kentaylor 

Having an app for configuration greatly improves the user experience while using the library. The cost is of course the additional code to allow programmatic configuration (e.g. serving JSON as well as HTML). I feel in this case, and this is totally my opinion, that the gain in user experience is much higher than the cost in the additional code and resources to have the programmatic configuration APIs.

That said, perhaps @tablatronix is right, tying the WifiManager to the app closely might not be what the WifiManager community wants, especially when the app would only serve a subset of the user base.

As a compromise, how would you feel about having a list of apps instead of just this particular single app? I feel that without adding a reference to the apps in the HTML pages that the ESP device serves, many of the users would not be aware of the possibility of having a much better experience via the apps.

BTW, please try out the app. Any feedback you can give me will be highly appreciated. It's available on [Google Play](https://play.google.com/store/apps/details?id=au.com.umranium.espconnect) and the code base resides [here](https://gitlab.com/umranium/EspConnect).

tzapu:
@tablatronix i m going through them manually because @kentaylor 's branch is heavily oriented on 'on demand' functionality and more expert uses. this is happening in the development branch
the kentaylor pull request branch is so i can mark his pull request as accepted at the end, i want to make sure contributors are given credit through all means available
where i hope to go with this is:
- have the json api, that works with the apps (but it is totally optional)
- have a list of apps and notable branches in the docs for people that want to experiment
- have a better interface that you can upload to your FS (also totally optional)
- rework the connection algorithm so it s more like ken tested and suggested (biggest point being that if it s set to auto connect, or if it connects in the background, it should do the right things, rather than force a connection)
- rework the documentation to give a more through overview of how you can use it in the two possible ways (auto/ondemand) and to have a paragraph explaining what you can find in each example
- ken also had a very interesting change of having a named host for the esp, which is a real domain, which could give more information and that is a very smart idea

each of these changes are potentially breakign and i would like to give people a chance to try them in the development branch as well before making it a release

my goal for this is to have it dead simple to use for everyone, paste and go. more experienced users will always find ways of changing things, but people just starting up should not need more than 3 lines of code to get started. if they want, they can explore further at any point.

also the fact that i have quite a few modules that have been working fine with power cuts and internet cuts and so on doesn't motivate me enough to have major changes to the 'philosophy' (big word, i know) behind it all :D 

thanks for letting me know it was featured, had no idea :D lovely

thank you guys for everything, you are a wonderful community :)

tablatronix:
Sounds good, but I think the main feature of this lib is its simplicity and compactness, it should be very small, since it takes memory and is hardly ever used, which is very important. ( I really would like to the the new styles brought it immediately, the notification blocks, the big buttons, those minor css changes are very nice, and my icons of course )

I am also thinking maybe all this on demand stuff is over complicating it, all we really need is for it to detect that wifi is not working ( end user/developer decides what this means ,no ssid or no ap or no internet ping, or no ip etc ) . The simplest and most obvious solution is the following.

* Auto start portal if no SSID previously setup, user can run this detection whenever they want.
* Erase network config ( possibly keep other fields ) ( which forces above condition )

Thats it, real simple, even for end users, start config if no ssid, clear ssid, 2 steps and any user can provide config portal and reset mechanism, being hold button on power up, hold button for 10 seconds, cycle 3 times in 5 seconds.

Yes there are many advanced options available
on demand portal, for using to config extra fields at any time
persistance portal as a status page
but these are all easily controlled by end users with easy examples.

So we are looking at a few methods

autoconnect
configportalstart
configporttalstop
erase configportal settings ( some/all)

clean responsive UI
customizable templates all that jazz

But if you support extra fields, then ondemand needs to be available easily
If you support using status pages ( pretty cool plus ) people can use this to throw up real fast simple iot devices, eg just a temperature field, etc. ) and this is just a AP+STA setup so not really to much extra code

The saving and handing off is a definite problem, if you are looking at professional or commercial design then yes this needs to be well thought out, but if there is a standard for this then cool, use apps, but we have no idea who makes these apps, what they do whats in them. 

I think that is kens approach, an app or a real url, this is how alot of iot devices allow config, my thermostat, my scale all use this method, and it to keep it available when it changes from ap to sta and networks switch and drop when restarting, typically you just use js loop and redirect to the new ip after reboot, but esp presents the problems ken adressed, wifi channel switching and all those tiny issue that cause sta+ap+scanning to cause drops in service during setup.

I am not sure there IS another solution.

tablatronix:
Just curious have you experimented with mdns for local urls as a solution ?
esp8266.local for example ?

tzapu:
intially it was based on mdns, never worked that good back then, i kept having to use the ip anyway
after dns redirects came, switched to that. i see ken isusing a specific hostname.

i shall do more tests with jsloops, there must be  a way in which that will work fairly well, although not for captive portal which closes when the wifi point goes away... (it just increases the code size for a one off thingy)

tablatronix:
Ken mentions some of the login portals will timeout and close while the config is saving and conx is lost. Ios etc. so maybe the real hostname keeps that page up while it switches aps not sure though seems solveable if you used a browser instead of auto launch login page. Just a thought. I do not fully understand all the changes

Daemach:
I want to use this library to set up wifi, then redirect the user to a
public website, passing the unit's chipID + mac as a serial number so the
user can register the device, attaching it to their account for
monitoring.  What's the best way to accomplish this?

tablatronix:
Ask questions in a new issue, this is a pull request discussion.

tablatronix:
minor annoyance.
My compiler throws this, 
`WiFiManager.cpp:349:23: warning: deprecated conversion from string constant to 'char*' [-Wwrite-strings]`
returning strings function declared with char*

kentaylor:
@Daemach the short answer is try [IOT Configurator](http://configure.urremote.com/) and while @tablatronix  would prefer this to be a new issue it can be interpreted as a prompt to consider the functionality WiFiManager should provide. 

I realise this is left field but I think the @Daemach preference for obtaining custom application parameters from an online service is likely, over time, to become the default and accordingly consideration should be given to keeping WiFiManager minimalist and splitting out the custom parameter functionality to a separate library.

Reasoning:- 

1. Configuring WiFi is a tightly defined problem that is amenable to a universal solution.
2. Getting a WiFi connection has to be done locally and is a separate problem to custom parameter configuration which can be done differently.
3. Custom parameter collection and custom display is, in it's most generic form, addressed by a web server and when made more specific is akin to a [web application framework
](https://en.wikipedia.org/wiki/Web_framework).
4. A web application framework type library is useful for all situations where a web interface is desired, not just WiFi set up.

kentaylor:
I've been intending to incorporate some items from this discussion in the https://github.com/kentaylor/WiFiManager version. 

I liked the simplicity that could be achieved from the @tablatronix [suggestion that](https://github.com/tzapu/WiFiManager/pull/157#issuecomment-260187887)

- Auto start portal if no SSID previously setup, user can run this detection whenever they want.
- Erase network config ( possibly keep other fields ) ( which forces above condition )

but when looking at how to implement it, noticed that it would conflict with the [Wishlist item](https://github.com/tzapu/WiFiManager/blob/60c22d672fadfdcf139853f766d1824730944e13/README.md#wishlist) add multiple sets of network credentials and particularly [this algorithm](https://github.com/tzapu/WiFiManager/issues/166#issuecomment-218134202). 

Therefore the enthusiasm for that item has waned.

tablatronix:
I see,  so implement wifi multi ?
hmm, yeah that does cause an issue with that reset flow , although I doubt your typical IOT appliance user actually needs this functionality, especially if it is so easy to change with wifimanager.

How often would be a device that moves around be needed or what else might a user need that for.

kentaylor:
@tablatronix I move ESP devices among a few networks quite often and regard reconfiguring each time to be a pain. I can't think of other reasons multiple SSID's would be useful.

IOT devices aren't usually moved as often as phones or laptops but imagine the annoyance if you had to reconfigure phones and laptops each time you moved between networks.

Daemach:
Thanks - that was my thinking as well.

tzapu:
guys, maybe another way to think of the multiple networks support is to have them in local storage on the phone ... a little bit of js, and not as convenient as auto switching from network to network, but gives you more control. i tend to also use the same device to reconfigure devices, so at least for me it would have instant access to the previous selected networks.
as opposed to ken's usage, there are people that have multiple networks in the same area and connect the device to each of them in turn, having all networks saved and connect automatically to one of them (not being 100% sure which one) would probably be quite annoying...

kentaylor:
The @tzapu scenario does not conflict with automatic multiple network connectivity. To stick with the laptop/phone analogy, it is equivalent to clicking forget network. 

Like some laptop/phones you could be more sophisticated and have a preferred network. Maybe make the first network in the array preferred and an algorithm that said use this unless the signal strength is below a threshold.

The biggest negative in my mind is the effort required to implement it. WifiMulti is a good starting point but the interface would be a pain. It has to support adding and removing APs like a laptop/phone does. Looking a bit at the Arduino library, WiFI.begin doesn't support it either so that has to be changed or else call the Espressif functions directly.

tablatronix:
I agree that this is way more complex i was gonna post in the actual issue for it #166

PR #181, opened 2016-06-02
F/non blocking api

mrbig:
Hello,

this update introduces a non-blocking mode for the WiFiManager.

My devices use wifi merely to pull updates, and the don't even have connectivity every time they turn on. So blocking for several minutes with the config portal is not an option: I wanted to have my application run parallel with the config portal.

In this change I've broke down the main blocking functions (connect, portal event handling, portal connection)  into smaller ones, and introduced an internal state maching (_currentState attribute) to keep track of the state.

Default is still the blocking mode, to provide backward compatibility with the original implementation. But in the internals it will do the non-blocking style, but wait before returning.

I've tried to keep the changeset minimal, unfortunately I had modify several functions.

There are still some small blocking events like the scan process, yet I believe they usually don't take more than a few seconds.

There's an example included to show how non-blocking mode can be used.

Any comments and suggestions are welcome.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -79,11 +79,11 @@ void WiFiManager::setupConfigPortal() {
 
   DEBUG_WM(F("Configuring access point... "));
   DEBUG_WM(_apName);
-  if (_apPassword != NULL) {
-    if (strlen(_apPassword) < 8 || strlen(_apPassword) > 63) {
+  if (_apPassword != "") {
+    if (_apPassword.length() < 8 || _apPassword.length() > 63) {
       // fail passphrase to short or long!
       DEBUG_WM(F("Invalid AccessPoint password. Ignoring"));
-      _apPassword = NULL;
+      _apPassword = "";
     }
     DEBUG_WM(_apPassword);
   }
@@ -94,10 +94,10 @@ void WiFiManager::setupConfigPortal() {
     WiFi.softAPConfig(_ap_static_ip, _ap_static_gw, _ap_static_sn);
   }
 
-  if (_apPassword != NULL) {
-    WiFi.softAP(_apName, _apPassword);//password option
+  if (_apPassword != "") {
+    WiFi.softAP(_apName.c_str(), _apPassword.c_str());//password option
   } else {
-    WiFi.softAP(_apName);
+    WiFi.softAP(_apName.c_str());
   }
 
   delay(500); // Without delay I've seen the IP address blank
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -145,7 +145,11 @@ boolean WiFiManager::autoConnect(char const *apName, char const *apPassword) {
   currentState = WIFIMANAGER_STATE_CONNECTING;
   if (connectWifi("", "") == WL_CONNECTED) {
     afterConnected();
+    return true;
   }
+
+  if (!_nonBlocking) return block();
+
   return true;
 
 }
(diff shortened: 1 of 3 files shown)


mrbig:
Hello,
I just realized that one of my examples did break the auto-build, so I've fixed it.
Will you have some time to check this update, or does it conflict with some of your plans?

Htank you

Jorgen-VikingGod:
@mrbig I tryed your fork and worked great.
But how do I use WiFiManagerParameter in combination with non blocking?

tzapu:
good to know guys and thank you very much,
again, i am sorry for not being able to do more these days, but hope to catch up, hopefully soon

cheers

mrbig:
@Jorgen-VikingGod it should work as with the default. I didn't change the way the server handles requests, only split up the wait cycles into callbacks.
Let me know, if you had success with the WiFiManagerParameter as well.

Jorgen-VikingGod:
@mrbig To use WiFiManagerParameter, I have to create them global - in the case I want to load settings from flash and fill in the defaults. And the other way - if config is closed, I have to get values from Parameters and store them in my variables and on flash storage.

I always got dirty stack overflows...

Don't know, if it depends on the WiFiManagerParameter technique or on the blocking part...

I will try this a bit and report again. 
Currently you can only add parameter by
https://github.com/mrbig/WiFiManager/blob/f/non_blocking_api/WiFiManager.cpp#L66
but no way to get parameter depending on the id...

Maybe I have to create a new function to WiFiManager to return Parameter object...

tablatronix:
This seem like a nice feature, but i am wondering if it is not easier to just require user code and intervention to enter config mode at will, so as to not bloat the code, i can also see this creating another point of failure into peoples code.

mrbig:
Hi tablatronix,
nut sure I understand your point. With this the default behavior did not change: if you keep your original setup the lib will keep working as usual. On the other hand if one needs the non-blocking behavior then the only change required is to put a call in the main loop. This is pretty much standard with such behavior.
My problem with the original wifimanager is, that it blocks every other code execution. In my usecase I have to execute my standard code even in config mode.

tablatronix:
I was not arguing against the features usability, just whether or not this pull request should be done or not when comparing to other PR , it might introduce a bunch of refactoring.

I see you added a manual processor
wifiManager.process();

I do like this approach now that I look at it.

mrbig:
I understand your point. It's a major change, with all it's risk. To try to minimalize that I've added a test as well, and made sure that all the previous tests work well.
I'd be happy if others would try this branch in their projects, so I can fix bugs if any exists.

copercini:
for ESP32 it will be very easy using WiFiEvent, like:

wifiDisconnectedLoop(){ open portal }
wifiConnectedLoop(){ close portal }

based on this example: https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/examples/WiFiClientEvents/WiFiClientEvents.ino

MarcFinns:
This is a very useful enhancement! Is it going to be incorporated at some point? It was proposed 8 months ago and I am losing hope...

tablatronix:
I have implemented some of this feature in my dev PR, I refactored some stuff , and did not see the need to implementing all the stated based stepping as it was over complicated, plus it made the code to convoluted to be able to merge other PRs in.

mrbig:
Hello @tablatronix can you please point me to your PR? I would like to check out how much does it fit to my needs.
I'm not sticking to this PR, if non-blocking operation is available in some other way.

tablatronix:
https://github.com/tzapu/WiFiManager/pull/313

tablatronix:
It needs a review and testing there might something I overlooked as it only handles the configportal processing and does no timeout handling for the user etc.

PR #188, opened 2016-07-02
Fix to allow 5 digit 'mqtt_port'

bakrus:
'WiFiManagerParameter custom_mqtt_port' made consistent with declaration of 'mqtt_port'


tzapu:
thanks

PR #240, opened 2016-10-21
simple function pointer to determine when to exit config portal

mvkarthikeyan:
please tell me if i am doing anything wrong coz i am newbie in C++


tablatronix:
can you provide an example scenario ?

mvkarthikeyan:
In our Project we need to exit config portal using a button click on NodeMCU .

PR #244, opened 2016-11-15
several fixes for max length of parameters

black-snake:
This is to fix the problem that you have to initialize a custom
WiFiManagerParameter with a length that is one more than you actually
intend. Also, this fix limits the HTML input fields to the maximum
length defined for the respective WiFiManagerParameter. Generally, the
length of a custom WiFiManagerParameter is currently limited to two
digits (=99). In addition, this fix includes a limitation of the HTML
input fields for WiFi SSID and WiFi password to the maximum length
allowed.


tablatronix:
there was another PR that already merged

PR #251, opened 2016-11-26
Adding WiFiManagerParameter implementation with labelPlacement parameter

ayomawdb:
Without the suggested change, below error will occur during compilation (if parameter with custom "labelPlacement" was added): 

`undefined reference to WiFiManagerParameter::WiFiManagerParameter(char const*, char const*, char const*, int, char const*, int)`


tablatronix:
Not sure what this means, is this bug still present ?

PR #282, opened 2017-01-07
translation support added, web generation improved

thewilli:
Hi!

I am using your great WiFiManager with some people who do not understand English, so I hacked some translation into it, which didn't work well. So I decided to extend the project to support different languages (one at a time). To be less error-prone to web modifications I also optimized the web workflow (replaced it by a [webpack](https://webpack.github.io) one) and added support for template parameters, so you don't need to do String replacement anymore :wink:

All changes are mentioned and explained in the README and corresponding commit message.

Best regards,

Willi


tzapu:
all i can say is .... WOW
great job man, it will take me some time to go through it, but really impressive stuff

thanks for this

thewilli:
@tzapu great you like it 😃 

Any news regarding your review?

tablatronix:
What is the nojs fallback for this ?

Can the i18n be abstracted out? It looks like yes, you simply moved strings to include, and added output wrappers from what i see at first glance.

laurent7878:
hi, any upgrade?

PR #284, opened 2017-01-07
use setAPStaticIPConfig instead  of (old?) setAPConfig in first example

robertoostenveld:
The (old?) function call is only done in a commented-out section, so the bug does not directly affect the functionality of the example code. But I encountered the error when I started playing with the various options of the WiFi Manager.


tzapu:
thanks for this

Issue #285, opened 2017-01-08
When WIFI_MANAGER_MAX_PARAMS is too small the library messes up previously defined custom parameters


vyruz1986:
By default this is set to 10.
In my project I had to define more than 10 custom parameters, but was not aware of this limitation.
The problem is that even when the `WIFI_MANAGER_MAX_PARAMS` is exceeded the library keeps happily adding extra parameters, but is actually overwriting previously added ones. Probably due to the array being only 10 large it wraps around.

It would be better in `WiFiManager::addParameter()` to check what the current `_paramsCount` value is, and when it is higher then `WIFI_MANAGER_MAX_PARAMS`, discard the new parameter and make a debug logging inform the developer that he should increase the define.

PR #286, opened 2017-01-08
Don't add custom parameters if array size is exceeded

vyruz1986:
array containing custom parameters is initialised at class
initialization time with a fixed value. When we later try to add more
parameters than the array allows we will now discard the offending
parameters and make a debug logging so the developer knows what's going
on.

Might be further improved as per this answer on StackOverflow:
http://stackoverflow.com/a/12271501
This would allow the array size to be specified at initialisation time,
and would give flexibility to the developer to decide how big the array
should be.

closes tzapu/WiFiManager#285


tzapu:
thank you very much

Issue #308, opened 2017-02-07
configPortalTimeout improvements


tablatronix:
only timeout configportal if ap has no client

tablatronix:
pr 
https://github.com/tzapu/WiFiManager/pull/310

tablatronix:
My fix was simple and to keep configportal open if a client was connected to AP.

This avoids dealing with last request timestamp bumping and all the problems that creates and fails to solve for.

tzapu:
in my mind, the timeout should be paused or reset if there is a client connected to the ap or maybe just increase the timeout interval with 30 seconds (for example) each time the loop happens, it has not timed out yet and there is someone connected... this should give in effect a timeout that will happen 30 seconds + whatever was left of the timeout after the client disconnected...

a bit hacky maybe, but not bloated i think

tablatronix:
So just add a static global timeoutbuffer variable and add it to the timeout calculations ?

tablatronix:
hows this?

```
boolean WiFiManager::configPortalHasTimeout(){
    if(_configPortalTimeout == 0 || wifi_softap_get_station_num() > 0){
      _configPortalStart = millis(); // kludge, bump configportal start time to skew timeouts
      return false;
    }
    return (millis() > _configPortalStart + _configPortalTimeout);
}
```

tzapu:
that looks brilliant to me, thank you very much

tablatronix:
pushed to PR

tablatronix:
In the future maybe implement a keep alive headers or a microajax keepalive

tablatronix:
It looks there might be a problem with this solution.

Depending on how the client disconnects, wifi_softap_get_station_num continues to report client connected.

tablatronix:
This seems to be caused as with a ton of other problems, by bad sta connect while trying to run softap, it seems it causes issues with dhcp as well.

I was able to stabilize this a bit by `WiFi.disconnect(true);` before enabling softap if not connected.
setting `mode(WIFI_AP)` was not enough.

FernandoGarcia:
Hi @tablatronix @tzapu!

After this update the portal never goes to timeout even without any client.

Best regards.

FernandoGarcia:
I'm sorry was my mistake I don't know the reason but I changed the timeout from seconds to milliseconds. :D

tablatronix:
added webclient checking to bump timeouts, good enough until I decide to add js and ajax keepalives.

[...]

PR #310, opened 2017-02-08
Feature 308 configportaltimeouts

tablatronix:


tzapu:
thanks for this, 
if i read it correctly ,will this exit the flow as soon as the connected party disconnects?
would it be more intuitive to reset the timeout if someone is connected? or make it longer?
just thinking outloud, but if you connect with your phone, then decide to use your laptop to configure, disconnect from phone, it will probably restart by the time you connect with the laptop.
although, not sure i read it right

cheers

tablatronix:
If the timeout was set and has exceeded, yes. Right now it timesout when you are using the portal which is not great.

If a webclient timeout is required, then I think it should be a seperate feature, with a bumped timestamp on every request. I thought about using configportalstarttime but that seemed hacky we would want a new timestamp configportallastactivity for example, so decided it should be 2 seperate things, as the later requires new functionality and globals. And there is some decisions to be made about how that would work as it could also timeout when you are still typing in configuration forms etc.

tablatronix:
note, I updated comment in case you are using email...

tablatronix:
I reopened #308 to discuss, since this might cause unexpected changes

PR #318, opened 2017-02-18
Add "Exit" button option to main webpage

Rom3oDelta7:
Add a new "Exit" button to the main webpage to allow the user to exit without saving WiFi credentials. One reason for this would be for the application to put the ESP in AP mode if WiFi credentials are not provided; this allows an application providing HTTP services to work either standalone or on a local WiFi network.


tablatronix:
implemented in dev, copied your readme

PR #322, opened 2017-02-24
Custom api endpoint

Jimut:
Added options to add custom API endpoint when the configuration portal is opened. 

Please take a look at it. I think it will help people who have an unreliable network connection, during such times they could read and write data directly from the ESP server.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -63,6 +63,16 @@ const char* WiFiManagerParameter::getCustomHTML() {
 WiFiManager::WiFiManager() {
 }
 
+//store the api endpoints
+void WiFiManager::addAPIEndpoint(char const *endpoint, void (*callback)(WiFiManager* wifiManager)) {
+  _endpoint[_endpointCount] = endpoint;
+  _endpointCallback[_endpointCount] = callback;
+  _endpointCount++;
+
+  DEBUG_WM("Adding endpoint");
+  DEBUG_WM(endpoint);
+}
+
 void WiFiManager::addParameter(WiFiManagerParameter *p) {
   if(_paramsCount + 1 > WIFI_MANAGER_MAX_PARAMS)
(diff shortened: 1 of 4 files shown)


Jimut:
Sorry for this, but I don't know anything about Travis CI or why it is failing. I have to learn it.

aj-ptw:
this is awesome

tablatronix:
Is there a better way to do this? Like expose the webserver itself, and let user code add their own routes and callbacks , I do like this implementation, but does it even need to exist ?

Or instead of adding routes to WM, can the webserver library catch all or wildcard, and we can simple hand those off to a callback with a single closure ?

aj-ptw:
@tablatronix the problem for me is my web server implementation is too memory intensive and causes crashes if i start the WiFiManager post web server.

tablatronix:
looks like no wildcard or subtree handling exists, it only does uri equality comparisons, so that is not helpful. notfound still looks like a good option, if it can pass the requestobject along.
```
        if (requestUri != _uri)
            return false;
```

aj-ptw:
>  i just mean not have to use WM for handling the routes internally, more code to maintain, etc, plus if you port to async websever or other webservers, it causes issues.

So if we look inside WM we see that it declares it's [own private web server](https://github.com/tzapu/WiFiManager/blob/89af31245b98b05e7d89319746ef3bc29d7e965b/WiFiManager.h#L115) which does not make this a light weight library. Perhaps we can pass our own webserver into WM to prevent WM from automatically spinning up it's own web server.

tablatronix:
That is what i was thinking, passing in reference or pointer to webserver, or somehow expose the wm webserver or even both.

tablatronix:
Why did you add an endpoint max and tracking bucket? Do you think that is necessary, do we need it ?

PR #365, opened 2017-05-04
Allow showing static IP config without predefined value

vyruz1986:
Previously a valid predefined static IP had to be passed before the
config page would show the IP, gateway and subnet fields. This did not
allow the config page to show 'optional' static IP fields.
Now a boolean can be passed to the autoConnect() function which will
specify whether the static IP fields should be shown or not.
When this boolean is true but no predefined values are passed and fields
are left blank in config page, wifiManager will revert to DHCP. This
allows user to choose between static IP or DHCP without changing code

For backwards compatibility we will show the static IP fields regardless of how the boolean is set.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -500,7 +500,7 @@ void WiFiManager::handleWifi(boolean scan) {
     page += "<br/>";
   }
 
-  if (_sta_show_static_fields) {
+  if (_sta_show_static_fields || _sta_static_ip) {
 
     String item = FPSTR(HTTP_FORM_PARAM);
     item.replace("{i}", "ip");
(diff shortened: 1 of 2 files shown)


tablatronix:
I think the argument should be dropped, and replaced with a setter for this option. We probably do not want to clog up the argument list with flags or options.

vyruz1986:
That makes sense to me, are you going to implement this or do you want me to do it?

tablatronix:
I can tweak it after pulling it in

tablatronix:
hmm ill have to cherry pick it into development

tablatronix:
oh nm you can now do it from the web !

PR #378, opened 2017-06-26
Add a callback when reset settings is triggered

gmag11:
I've been using WifiManager library. It is such a great tool, thanx,

When reset settings I would expect that all custom settings are deleted too. To do that I've added a callback to notify main program about this event so that it can do custom reset settings task.

Decide if it is a good feature to be included in your repository.


tablatronix:
#452

tablatronix:
- [ ] change cb to std function

PR #387, opened 2017-07-13
Add bool autoConnect() function

JXproject:
Although autoConnect() can directly connect wifi, but it turns into AP mode right away if it can not connect to the wifi.
And this will be super limited to users. In some cases, for example, user want to know if they can connect to wifi, and go for sleep   ##after several attempts and avoid directly entering AP mode. Only enter AP mode by some hard reset, which can increase the network security when power outage happened with device/network.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -157,6 +157,23 @@ boolean WiFiManager::autoConnect(char const *apName, char const *apPassword) {
   return startConfigPortal(apName, apPassword);
 }
 
+boolean WiFiManager::ConnectWifi(){
+    DEBUG_WM(F(""));
+    DEBUG_WM(F("Attempt to Connect Wifi"));
+
+    // attempt to connect;
+    WiFi.mode(WIFI_STA);
+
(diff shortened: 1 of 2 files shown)


dronov:
:+1:

tablatronix:
I thought about adding this, but the esp will auto connect on its own in the background. So if you do not want the config portal just don't use auto connect at all. 

However, I can see the need , someone might be using WiFi.setAutoReconnect ( false ) and manually starting up wifi when they want to, in this case autoconnect might help, of course you could also just call begin in code.

Still on the fench

tzapu:
i m thinking we should change autoConnect to not call begin at all if auto reconnect is true.
this should sort a load of other problems with forcing the connection.
it should probably wait until a timeout.
if auto reconnect is false, it should maybe throw a warning and force connection with begin

this is more along the lines of working with the SDK rather than against it, that i mentioned...

tablatronix:
Revisit adding a autoconnect loop before calling begin

PR #411, opened 2017-08-31
multiple network support

the-real-orca:
I added some code to handle multiple networks and credentials.
If the connection to the last connected network failed, a list of known networks will be searched for the strongest AP in reach. The known networks can be load and stored to the internal FS (see examlpes/AutoConnectMultipleNetworksWithFS).
It will not re-scan or reconnect to another network when the connection is lost. WiFiMulti is not used (even if the branch name would suggest).


the-real-orca:
While updating and rebasing to the development branch I messed up my repro and also accidentally deleted the remote branch. Now everything is on track again and I re-submitted the patch against the development branch. -> #622

PR #447, opened 2017-10-31
Proposal for various form inputs

Thoma-HauC:
Hi,

I needed various form inputs for a small project.
I modified your original design, therefore I removed WiFiManagerParameter and replaced it by an ancestor web_parameter.
I derived four classes : text_box_parameter, password_box_parameter, check_box_parameter and radio_field_parameter.
text_box_parameter is similar to WiFiManagerParameter.
Hope this will be helpful to other people.

Regards,

Thoma


tablatronix:
No pr against master

tablatronix:
Oh 4 months ago oops

tablatronix:
Whats webgets?

tablatronix:
This is pretty interesting, i will take a good look at it.

Thoma-HauC:
Hi Tablatronix,

"webgets" contains the definition of text input, password input, checkbox and radio button.
There is a common ancestor in order to store the different instances.

dontsovcmc:
Thanks!
It's better to add different data type TextBox parameters:
- String
- Integer
- Float
- Boolean (check box)
- IPAddress
- ComboBox

with getters & setters. 

The another feature: checking input values. If error - MessageBox or test label will showed.
Example:
```
bool x::handle_check(data_type &value, String &reason)
{
       if (value > 10 & value < 100) {
              return true;
       }
      reason = "value should be between 10 and 100";
      return false
}
```

dontsovcmc:
@Thoma-HauC, is it error in your examples:
Sometimes blynk_token parameter is 32, sometimes 34.
```
text_box_parameter custom_blynk_token("blynk", "blynk token", blynk_token, 32);
text_box_parameter custom_blynk_token("blynk", "blynk token", blynk_token, 34);
```
As blynk_token has 32 characters, we need to add '/0' terminator. Some string functions found of terminators in the character arrays.. )

tablatronix:
That is not the case in development branch

dontsovcmc:
@tablatronix Different parameters and check input? See my first message.

tablatronix:
Null terminator extra length

tablatronix:
Should these classes be sub classes of wifimanager?

Can they be used separately ?


I would love to see this syntax simplified somehow, some sugar.
```
text_box_parameter custom_mqtt_server("server", "mqtt server", mqtt_server, 40);
wifiManager.addParameter(custom_blynk_token);
strcpy(blynk_token, custom_blynk_token.get_value().c_str());
```

It is alot to ask users to do to just add a form..
Needs to be abstracted a bit more.

It would probably require storing the params in the class, so we can add a 
register param(id,options)
get param(id,@ref)

or something

PR #456, opened 2017-11-14
Connection Initialization Fixes

maherahmed:
To solve the problem occurs after a power cut (WiFi router takes a while to broadcast a WiFi network, so esp8266 searches for the saved network but of course it will not be found, so switches to ap_sta mode and wait for an external config via the captive portal) -> solved by  making the esp8266 keep trying to connect using last saved values even if it is not found at the start of autoConnect() and if it connected get out of autoConnect()  Edit#1

To solve the problem of connected but still waiting for the ssid and password in the same time edit#2,3

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -196,7 +196,17 @@ boolean  WiFiManager::startConfigPortal(char const *apName, char const *apPasswo
     //HTTP
     server->handleClient();
 
-
+	  //added to solve the problem of power cut -> keep trying to connect using last saved values edit#1
+	  
+	if (WiFi.status() == WL_CONNECTED){
+	WiFi.mode(WIFI_STA);
+	DEBUG_WM(F("connected with last saved values, AP will be closed"));
+	DEBUG_WM(F("IP Address:"));
+        DEBUG_WM(WiFi.localIP());
+        break;
+	}
+	 //edit#1
+	  
     if (connect) {
       connect = false;
       delay(2000);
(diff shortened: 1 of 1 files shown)


tablatronix:
This would cause several problems with on demand configportals, also whynot just add a delay to your sketch before autoconnect, or increase autoconnect timeout?

maherahmed:
@tablatronix I do not think a delay would be a right solution to the problem. I'll see the issue of on demand configportal.

maherahmed:
@tablatronix could you point out what are the problems with on demand configportal ?

EUA:
This patch has to be included. Saved my long term problem.

Adding `delays` doesn't work because, this device could powered with in-depend power source (for example feed by another phase at building as I have problems here for a while.) 

So please include this patch.
Thanks for the solution.

tablatronix:
You cannot leave sta mode on when starting an ap if its not connected, it breaks the configportal.

I still do not unserstand what the problem here is, just increase the connecttimeout or add delays in your srartup code or check for connection before calling autoconnect.

tablatronix:
alternativly use configportal timeout and reboot if still not connected afterwards

tablatronix:
or set an event of your own and stopconfigportal, but its worth noting that sta kode is ALWAYS disabled in hotfixes and development branch now.

tablatronix:
Add an option to auto close autoconnect portal if background connection is established. I imagine most people will use non blocking now anyway

PR #460, opened 2017-11-19
FIX: Server was private, now public

andrewjaykeller:
If we expose the server publicly, then we can use it outside of WM and save on memory footprints! In favor of #322


tablatronix:
#461 
thoughts on ownership, memory usage etc welcome

tablatronix:
This is already done in dev

PR #467, opened 2017-12-05
Document that parameter id cannot contain spaces

lukepalmer:
It took me a while to figure out that custom parameters don't work if there is a space in the id parameter. It's not particularly useful to make this work, so just document that it doesn't.

The reason it doesn't work: the client will use the id parameter as the name of the field in the HTTP GET request that goes back to the server. Spaces and other special characters will get encoded, so "Blynk token" becomes "Blynk+token". The server doesn't do any kind of decoding so it gets confused and thinks the parameter just isn't there.

### DIFF
--- a/WiFiManager.h
+++ b/WiFiManager.h
@@ -39,6 +39,10 @@ const char HTTP_END[] PROGMEM             = "</div></body></html>";
 
 class WiFiManagerParameter {
   public:
+    /** 
+        Create custom parameters that can be added to the WiFiManager setup web page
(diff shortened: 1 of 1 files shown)


tablatronix:
* [ ] add encode decode, or enforce legal characters

tablatronix:
e5c263c4146d8b1f4f85ac1e73ca293709815f41

tablatronix:
addparameter now return bool, and it returns false if param is no alpha, although I can reverse this in the future or make optional, and users can use [I} token to use param_n form ids instead.

Issue #415, opened 2017-09-08
feature templating 


tablatronix:
I would like to add basic templating , ideas on how to implement welcome.

For starters, modifying http_input string for scan rows, what is the best way ?
A callout on wifiscan, a setter, some other templating method

tablatronix:
adding wifiscan template tags
{R} RSSI
{e} encryption type

tablatronix:
Adding some styles
adding callouts,  similar to kentaylors branch


![chrome_2017-09-08_13-55-23](https://user-images.githubusercontent.com/807787/30226961-8425afd6-949d-11e7-9d31-713698293f85.png)

tzapu:
as a mention, there was a very nice pull request around here, using webpack to spit out all the templates needed in a header file. it was cool because it did all the minimising and uglyfying for you.

tablatronix:
I will look around, in only remember the json one but it was bloated with a personal project and jquery

tzapu:
i think this is the one
https://github.com/tzapu/WiFiManager/pull/282

thanks for all your great work

tablatronix:
Ill look it over, ignored that because I only say the translation part

jakerabid:
Would it help if I was to see if I can resolve the merge conflicts in #282 and do some testing? I suppose I could fork this and add the code I need, but that is a code deadend and I'd rather my time be useful :)

tablatronix:
This is on next milestone so not really

jakerabid:
okay, I'll fork and work against my own feature, and rework my sketch in the future once this has landed. Thanks for all your hard work @tablatronix

tablatronix:
Please post ideas also.
I thought about callout filters

jakerabid:
I'll have a think and see, but at the moment I feel like just having all the strings (along with their corresponding html) be parametrised and overridable on the server (IE on the esp8266) as in #393 would be the simplest method that would give us the smallest complied code.

I guess the other option would be to do the templating in javascript, using something like https://github.com/janl/mustache.js, or even better http://handlebarsjs.com/, then changing the content would be nice and contained.... We would be swapping processing power on the esp8266 for a larger download to the users phone but I am not sure that is really an issue.

tablatronix:
Well there is 2 things that need to happen all strings in a new include and callouts for real time logic

tablatronix:
[...]

PR #393, opened 2017-08-07
Add custom_saved_elemend to customize HTTP_SAVED block

dronov:
Hi everyone!
A few days ago I got a little task to do a more detailed instruction for users what to do with esp8266 at HTTP_SAVED html block. So I decided to make a function like setCustomHeadElement for it. 

Please tell me how can I make it more elegant way? :) Otherwise I think this little commit would be useful for somebody.

PR #474, opened 2017-12-19
Added the ability to change the text shown on the wifisaved screen

jakerabid:
I'd like to be able to change the text shown on the wifisaved screen, here is my attempt -is this a good approach?


dronov:
#393

jakerabid:
I do like your solution better @dronov - but neither of them are merged....

tablatronix:
#415

Issue #479, opened 2018-01-06
Add callback pure virtuals -discussion


rossbeazley:
Would you consider accepting a PR that adds support for overloaded "add callback" methods taking a pure virtual class, say, WifiManager:: ConfigurationCallback.

I'm in the position of wanting to using class instances (I don't think I'm the only one) to receive the messages back from WifiManager.

Happy new year btw

R

tablatronix:
I am open to it , also more callbacks

rossbeazley:
i've raised #482 as a starter for ten

tablatronix:
Can you add an example and post some scenarios pros etc

Issue #452, opened 2017-11-08
Callbacks hooks


tablatronix:
Feature callbacks needed requested suggested.

sidoh:
Hey @tablatronix!  Any idea if/when this feature will be merged?

I'm maintaining a fork of WiFiManager to support these features, and would be great to kill it if possible.

tablatronix:
I have not gotten any feedback on this issue or have a final plan on which to merge

sidoh:
Got it, thanks.  Any way to poke it?  #378 in particular seems like a very harmless change :)

tablatronix:
callouts for dynamic content
`httppagehead`
`httppagefoot`

PR #482, opened 2018-01-08
Add callback delegate support for AP and Settings callbacks

rossbeazley:
This is my first go at a PR to address #479,

im more interested in knowing if this kind of structure would be acceptable before I perform some manual testing and also check code style and memory management.

thanks for your time


tablatronix:
#452

PR #487, opened 2018-01-22
make WIFI_MANAGER_MAX_PARAMS customizable

mtiutiu:
I think it's pretty useful to have this parameter customizable so that users can change it from outside if required (e.g platformio, make, etc). 

For example when using platformio I can set this in platformio.ini: 
**build_flags = -DWIFI_MANAGER_MAX_PARAMS=20**

This allows me to change  the value at compile time without modifying the library source code. I also provided the default value for it when not defined outside or somewhere else - you can see  that in the code changes that I want to propose.

This approach can be used for other defines in this library source code if desired/required.

Very nice work btw regarding this library - very useful. Thanks.


tablatronix:
Note that setting this is no longer needed, as memory is autoallocated for params now

mtiutiu:
I noticed that. So yeah it kinda makes no sense now.

PR #499, opened 2018-01-31
Development

tablatronix:
tracking


tablatronix:
@tzapu 
ESP32 has been merged into development
I cloned master into hotfixes, so we can maintain it untouched, in case anyone wants to legacy maintain it.

I think I am going to merge development into master and make it default branch, thoughts?
Ill figure out versioning first I guess and check make sure none of the manifests are using master a release, and using actual tags.

aj-ptw:
Attempting to test with makeESPArduino as the build system and getting:

```
In file included from /Users/aj/Documents/Arduino/libraries/WiFiManager/WiFiManager.cpp:13:0:
/Users/aj/Documents/Arduino/libraries/WiFiManager/WiFiManager.h:16:25: fatal error: ESP8266WiFi.h: No such file or directory
compilation terminated.
```

Probably a define or something I need to include, digging deeper.

stsdc:
Thank You for porting this to ESP32 :heart:

Issue #408, opened 2017-08-30
params null terminated length fix


tablatronix:
https://github.com/tzapu/WiFiManager/pull/244/commits/18e2dd7c503317fbbf7845d9c01aec8e15e06903
black-snake

9160b5bd3b8b1974dc6a39af2ab629b09a1779e4
63dd714805c5e818b975fa773a3b549f5e7a884d

tablatronix:
alternative would be to strip whitespace

PR #505, opened 2018-02-06
Bug fix for problem discovered when setting length to 1 for additiona…

Loufylouf:
…l parameters

The initialization of the buffer holding the value stopped before the last character and storage did not take into account the extra-byte for end of string.

Bug is pretty easy to reproduce : try to use an additional parameter with a length of 1. This PR fixes it.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -34,7 +34,7 @@ void WiFiManagerParameter::init(const char *id, const char *placeholder, const c
   _placeholder = placeholder;
   _length = length;
   _value = new char[length + 1];
-  for (int i = 0; i < length; i++) {
+  for (int i = 0; i < length + 1; i++) {
     _value[i] = 0;
   }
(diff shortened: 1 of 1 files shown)


tablatronix:
did you test development branch >?

tablatronix:
#408

Loufylouf:
Seems like you missed the initialization of the buffer :)

tablatronix:
What do you mean?

Loufylouf:
I saw a lot of commits to take into consideration the extra-byte for the end of string. But I didn't see a commit fixing the one missing in the `WiFiManagerParameter::init` function.

tablatronix:
oh I see, thanks , it was a PR, I should have tested more.

hmmm, Do we need that loop if we are using strings?
memset or _value[0] = \0 instead?

I see a strcpy and tochararray on it and return it as const char*
sorry thinking outloud

PR #522, opened 2018-02-15
Callback during wifi setup loop (for example, to flash an LED)

chrismidgley:
This very small patch enables a new feature to provide for an optional callback during the wifi config.  It is optional, so non-disruptive to existing implementations.  If setSetupLoopCallback is passed a function, that function will be called prior to yield() inside startConfigPortal.  We use this to "fast flash" an LED to indicate (pin 2 on the NODEMCU) to indicate that the device is pending configuration, rather than in operational mode.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -764,7 +764,7 @@ void WiFiManager::setRemoveDuplicateAPs(boolean removeDuplicates) {
 }
 
 //callback in loop during the setup phase, to allow for third party execution during setup wait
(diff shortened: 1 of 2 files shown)


tablatronix:
#452

tablatronix:
development branch
```
//if this is set, portal will be blocking and wait until save or exit, is false user must manually `process()` to handle config portal
    void          setConfigPortalBlocking(boolean shouldBlock); 
```

PR #562, opened 2018-03-15
Added custom static DNS

debsahu:
Used vim in linux to edit the files. Hopefully didnt create a mess.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -586,8 +586,14 @@ uint8_t WiFiManager::connectWifi(String ssid, String pass) {
 
   // Setup static IP config if provided
   if (_sta_static_ip) {
-    WiFi.config(_sta_static_ip, _sta_static_gw, _sta_static_sn);
-    DEBUG_WM(DEBUG_VERBOSE,F("Custom STA IP/GW/Subnet:"),WiFi.localIP());
+    if(_sta_static_dns) {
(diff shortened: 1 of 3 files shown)


tablatronix:
I had free time so i fixed it, already merged, thanks

debsahu:
lol, thanks :}

tablatronix:
I did change it to always show on webpage though, if showstaticfields is on

debsahu:
learning to use github effectively, still a learning from googling.

tablatronix:
check out core.autocrlf = auto, it keeps whitespaces same as repo

debsahu:
So if the user only specifies ip, gw and sn, then the DNS will still pop up in the web interface. That's why I set it as optional, else we got to edit the readme that DNS is mandatory.

tablatronix:
Why would it be mandatory, it can be blank, the problem is showststicfields and if staticipmare tied togsther, i might add a more detailed customization option to flag each one regardless of custom or not

tablatronix:
Showstaticfields ip,gw,sn,dns bool, or maybe three state int 0,1,2 show always, show if set, show never

tablatronix:
Maybe just 2 toggles, ip and dns, if you enable staticip you probsbly always want gw and sn also always

debsahu:
Exactly, starting with and without dns.

PR #571, opened 2018-03-21
enable all example builds for esp32

liebman:
This shows that the esp32 spiffs not fixed.


tablatronix:
Is this still the case? I thought I fixed this already.

tablatronix:
ohh sorry misunderstood title.

Issue #461, opened 2017-11-19
shared webserver instance


tablatronix:
I am testing this already, but I am using a getter and a shared_ptr, not sure which way will be final.
Ill post my branch later, I was adding some callbacks and testing what kind of parameters and callouts we would want to for templating and sharing webserver control.

There are several options, make server public, add getter with shared_pointer, and af ew others.
I am unsure which is the best design pattern for this, or how we want to control ownership.

Thoughts welcome.

ageurtse:
Is there a way you could provide an example on how to run a shared webserver.

tablatronix:
Sure, when I figure out how

tablatronix:
dnsserver? dns server resolves domains

spilz87:
Hi I see 
As you can see I'm not a specialist ^^
You understand the idea : captive portal point to a "/captiveportal" page not to the "/" default page

spilz87:
An other idea could be to check if the client is connected to AP or from network , and link to different pages, if it is connected on AP the page as only few button which link to other specific pages, so the difference is only on the first page 
Do you see what I mean ? 
Maybe it's already what you do

tablatronix:
Yeah the idea is to share the webserver instance so you can access in your own code, you can do this now by just making it public, but I have not decided if that is the final route. It might be better to makde it a shared pointer so that memory gets cleaned up better.

spilz87:
A shared pointer allow you to keep the same webserver but the configuration of this webserver (.on(..) , etc.) is erase when captive portal is open, isn't it ? 
Do I understand well or I'm wrong ?
Let me know if it's not the good place to have this discussion

tablatronix:
I wouldn't think so, not sure what you mean. webportal is always on when configportal is on

spilz87:
Does the setupConfigPortal erase the setup of the server done in the setup() of the sketch ?

tablatronix:
Ah I see what you are saying, yes that resets the webserver. hmm you have to setup yours after that, wm would have to be setup to allow that to be skipped, it also gets torn down on closeconfigportal

tablatronix:
Curious , Did you add a new constructor to pass the webserver in , or use a setter?

I guess you could check by client ip, or get the clients attached to ap and check them for that ip request.

spilz87:
Not in WM, it was a test in a personal library 

Ok i will look for this function :)

spilz87:
If you remember the name, let me know ^^
Thanks

[...]

PR #587, opened 2018-04-09
WIP: Added OTA updater option from esp8266

sidey79:
Added OTA update support in config portal page


sidey79:
@tablatronix 
I Would like to know, if this is a feature you are willing to merge, before i put more work on it.
Otherwise, i can also provide an additional class which is a parent from your class.

tablatronix:
I doubt it needs to be in core. Unless its very lightweight

tablatronix:
I thought about refactoring menu and routes, so you can easily add menuoptions and routes, with callouts then you can just add this stuff fairly easily without modifying the lib.

sidey79:
The libs needed seems to be already included per default in esp8266 core.

But there are some additional strings needed for the pages. Not sure if this is lightweight enough.

If you provide callbacks i could separate the update stuff in another class. But it would need some access to the server object.

tablatronix:
#461

sidey79:
In general, the pull request works on eps8266. But i haven't tested it on an ESP32.

Im also not sure, what breaks the example tests.

liebman:
This is the error:

```
In file included from /home/travis/build/tzapu/WiFiManager/examples/AutoConnectWithStaticIP/AutoConnectWithStaticIP.ino:2:0:
/usr/local/share/arduino/libraries/WiFiManager/WiFiManager.h:46:25: fatal error: Updater.h: No such file or directory
```
Maybe that does not exist for the esp32 or has a different name?

sidey79:
I can't believe it...

https://github.com/espressif/arduino-esp32/tree/master/libraries/Update/src

`
Update.h	
Updater.cpp
`

liebman:
You should install the esp32 Arduino and test compile as there are some class differences.....

```
/usr/local/share/arduino/libraries/WiFiManager/WiFiManager.cpp: In member function 'void WiFiManager::handleUpdating()':
/usr/local/share/arduino/libraries/WiFiManager/WiFiManager.cpp:2419:3: error: 'stopAll' is not a member of 'WiFiUDP'
   WiFiUDP::stopAll();
   ^
/usr/local/share/arduino/libraries/WiFiManager/WiFiManager.cpp:2421:34: error: 'class EspClass' has no member named 'getFreeSketchSpace'
   uint32_t maxSketchSpace = (ESP.getFreeSketchSpace() - 0x1000) & 0xFFFFF000;
                                  ^
```

tablatronix:
I would file that as a bug with esp232 m thats messed up

tablatronix:
Merged into ota branch for testing.
i had to fix your urls they were backwards and didnt work.

I an probably going to pull this out into an example with user code insread of core, but this is working and testable, I added todos to code

tablatronix:
you need to fix your strings menu_ids it is missing the update token `_nummenutokens` index must match HTTP_PORTAL_MENU index

tablatronix:
https://github.com/tzapu/WiFiManager/commit/3e2f8ae9d81badcf00ca8d2c2cb8cb4c205e9cc1

sidey79:
sorry, i forgot to check this

sidey79:
I'm not sure what ist open to geht this also in the Development branch.

One thing which ist absolutely a must ist a paaaword protection i think.

tablatronix:
There is an ota branch for this, it works

tablatronix:
it has not been tested well and probably needs some minor improvements

Issue #500, opened 2018-02-02
🚧  Development Release /  Branch


tablatronix:
Development Branch
https://github.com/tzapu/WiFiManager/tree/development

**README NOT UPDATED**
- [ ] Create docs for changes, new features functions and known issues

tablatronix:
`setConfigPortalBlocking`
`setShowStaticFields`
`setCaptivePortalEnable`
`setRestorePersistent`

`startWebPortal`
`stopWebPortal`
`process`

Testato:
Very good,
for the esp32 merge i think it is fundamental and, at least for me, it work very well.
Merge it in the lib will help to have new test/opinion.

Testato:
- adds signal icons, replaces percentage ( has hover titles )

On this change i do not agree, imho have the percentage is better than a simple icons whit only 4 graphic step, and on mobile lcd you cannot click for see the percentage number.
It is a very important signal, example for decide the post orientation of the Esp on a critical position far away from the wifi-router.

You should add a public function that the user can use on the .ino sketch for decide himself witch type of dsta he need ?

I repeat, imho ot is a very important info that the user should be decide whitout change the lib code.

tablatronix:
include `<rom/rtc.h>` to get esp32 info to show reset reasons

dvukovic:
Has anyone tried WiFiManager with an ESP8285 ?

tablatronix:
I have one I have not tested yet

dvukovic:
To give a little background.
I am using the Arduino1.8.5 IDE on Win7-64 bit.
The code has been compiled for an NodeMCU Lolin v3 and works properly.
I have a ESP8285 m3 board that I have been working with and have interfaces an I2C OLED display and an I2C RTC chip.
I then compile the same demo programs as the NodeMCU, but the 8285 does not what too connect to any of my devices, ( DEL laptop, iPhone or Samsung tablet) as the NodeMCU has done.
Here is a screen shot of the IDE Tools configuration.
![esp8285 m3](https://user-images.githubusercontent.com/5108570/49023973-9dfe4100-f155-11e8-850b-8e6c02572393.png)
Please let me know if you see anything I missed.
Thank You for your time.

tablatronix:
What does serial debug say?

dvukovic:
After loading the program, the serial monitor shows:
[...]

Issue #441, opened 2017-10-18
example update


cwalger:
autoconnect example shows "setAPConfig" in comments but that function has been renamed to "setAPStaticIPConfig".  
Thanks for the hard work!

PR #596, opened 2018-04-20
Wifi Wait Callback

elliottkember:
When WifiManger is running the config portal, we currently don't have any control and can't make anything happen. We'd like to do some fun things to indicate current status, as well as escape the Wifi configuration step if we need to run offline. This change introduces a `wifiWaitCallback` that gets called every WiFi config portal loop.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -742,7 +742,7 @@ boolean WiFiManager::captivePortal() {
   return false;
 }
 
-void WifiManager::setWifiWaitCallback( void (*func)(void) ) {
+void WiFiManager::setWifiWaitCallback( void (*func)(void) ) {
   _wifiWaitCallback = func;
 }
 
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -742,7 +742,7 @@ boolean WiFiManager::captivePortal() {
   return false;
 }
 
-void WifiManager::setWifiWaitCallback( void (*func) WifiManager* myWifiManager) ) {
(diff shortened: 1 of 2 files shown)


elliottkember:
If the wifiWaitCallback returns `true`, we're exiting out of the AP setup to resume operation in offline mode.

tablatronix:
Just use configportal non blocking and call process in your loop

tablatronix:
although you can already do this 
https://github.com/tzapu/WiFiManager/issues/500

I might merge this so you have 2 options.. I just think people will do something stupid in this loop and wonder why their portal isn't responding properly.

elliottkember:
Oh, thanks - I didn't know about the non-blocking mode because it's not in the documentation. I can understand not wanting to merge to keep this foolproof - a warning not to do anything slow in the loop might do the trick. I'd rather not use development branch so for now I'm using my fork off master and it's working great. Very useful to have some visual feedback on wifi connection state either way!

tablatronix:
Yeah that issue is all that exists at the moment hopefully someone can help write some docs

elliottkember:
If I get some time this weekend to implement the branch and try it out, I’ll write up some usage docs. Is this branch close to ready / merging?

tablatronix:
Pretty much, there are about 2 major features that need completion, 90% of it is already documented in the overview in #500 and the source is well documented.

tablatronix:
oh and I need to fix up examples
#441

tablatronix:
Can you think of any reason to still require this?

elliottkember:
Not really — I switched to the asynchronous development branch and it's working great. Thanks!

PR #602, opened 2018-04-28
Fixing pointer causes repeated memory leaks

zenghi:
Repairing duplicate calls causes memory leaks

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -78,6 +78,7 @@ WiFiManager::~WiFiManager()
     {
         DEBUG_WM(F("freeing allocated params!"));
         free(_params);
+		_params = NULL;
     }
 }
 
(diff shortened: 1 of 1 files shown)


WiFiManager.cpp
```diff
@@ -78,6 +78,7 @@ WiFiManager::~WiFiManager()
     {
         DEBUG_WM(F("freeing allocated params!"));
         free(_params);
+        _params = NULL;
```
liebman: Given that this is in the destructor the memory containing the pointer is being freed as well leaving no risk of a double free. This is not needed.

tablatronix: I am going to merge it in case someone refactors code and moves this, it wont hurt.

tablatronix: oh master, nm, ill just add it to dev

PR #608, opened 2018-05-14
 add CORS headers

ohadcn:
this will allow external portal on the network to send credentials via browser.


tablatronix:
Should we make this optional or allow passing in parameters for source and access control ?

ohadcn:
in the rare case of someone connected to WiFiManager while browsing websites on internet this could be a security issue - as the website might send ESP wrong wifi credentials and make it useless.
except this rare case I see no reason to disable it.

but we can add it as parameter.

tablatronix:
Yeah it would be a very specific xss attack but who knows.

cbh6:
Any info about when will this be released ?

Thanks

PR #622, opened 2018-06-12
multiple network support (2)

the-real-orca:
I added some code to handle multiple networks and credentials.
If the connection to the last connected network failed, a list of known networks will be searched for the strongest AP in reach. The known networks can be load and stored to the internal FS (see examlpes/AutoConnectMultipleNetworksWithFS).
Unlike the ESP8266WiFiMulti it will not re-scan or reconnect to another network when the connection is lost. So, no need to run anything in the main loop.
(old PR: #411)


tackelua:
this feature is great, please keep it up

PR #645, opened 2018-07-07
disable wps by default, enable if its enabled in core

liebman:
In https://github.com/esp8266/Arduino/pull/4889 the core was modified to disable WPS by default.  Its only enabled if NO_EXTRA_4K_HEAP is defined.  To enable you must re-run the board generator with --noextra4kheap. This adds #ifdef NO_EXTRA_4K_HEAP around WPS usage so it won't generate compiler errors.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -654,6 +654,7 @@ uint8_t WiFiManager::connectWifi(String ssid, String pass) {
 
   DEBUG_WM(DEBUG_VERBOSE,F("Connection result:"),getWLStatusString(connRes));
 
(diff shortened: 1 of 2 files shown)


tablatronix:
We might want to apply this to hotfixes branch also

tablatronix:
I merged this without any esp32 exceptions, because I am pretty sure it needs work for that platform.

tablatronix:
#646

tablatronix:
hotfixes 
704a7ae393e8b8f820a73924d0480aae85781c26

PR #647, opened 2018-07-08
fix bad progmem access

liebman:
https://github.com/tzapu/WiFiManager/pull/608 introduced a new header located in PROGMEM but did not use FPSTR() in its use causing crashes.

Exception Cause: 3  [LoadStoreError: Processor internal physical address or data error during load or store]


tablatronix:
Thanks I didnt catch this.

PR #660, opened 2018-07-15
Fix crasahes in doParamSave() &  WiFiManagerParameter::setValue

liebman:
Had crashes in `doParamSave()`. 

Changed to use `server.hasParam()` instead of checking `server.arg()` for NULL, it returns a `String` not a pointer.

Don't use `.c_str()` when its being assigned to a `String`. I think using `.c_str()` on `String` that are transient is probably not safe.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -1228,9 +1228,13 @@ void WiFiManager::doParamSave(){
         break;
       }
       //read parameter from server
+      String name = (String)FPSTR(S_parampre)+(String)i;
       String value;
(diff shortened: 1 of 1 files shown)


liebman:
Wait on this - I think I'm still getting crashes :-(

liebman:
Ok so `WiFiManagerParameter::setValue()` used the length of the default parameter for buffer allocation.  In my case its for a webservice URL and I passed length as 1024 but default value as "" so it only allocated 1 byte.  Then when doParamSave() was called it copied a s string into that buffer trashing memory.

tablatronix:
Sorry about that, I was still testing this to catch some edge cases. I Should have used a branch.

tablatronix:
I clicked merge last night I guess It didnt go through oops

PR #680, opened 2018-07-28
Changed the year of copyright

jacienpang:
I have changed the year of copyright from 2015 to 2018.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2015 tzapu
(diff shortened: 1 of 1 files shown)


bertrik:
I don't see why the copyright date should be updated to 2018, if (parts of) it date back to 2015, I think you should maintain date as it is. See also
https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

tablatronix:
That is how copyright dates work you eother out he range or the latest date

Issue #678, opened 2018-07-27
custom html parameters


Laserlicht:
#### Hardware
WiFimanager Branch/Release:  Development

Esp8266/Esp32: 

Hardware:			ESP32

Core Version:      	staging

### Description

Bug in develop branch: custom html parameters aren't possible.
They aren't visible on page.

For example:
```
WiFiManagerParameter custom_text("<p>This is just a text paragraph</p>");
wifiManager.addParameter(&custom_text);
```

### Sketch
I have to comment out the following lines to solve the problem (WiFiManager.cpp).
Custom-HTML-Parameter don't have an ID.

```
  /*if(p->getID()){
    for (size_t i = 0; i < strlen(p->getID()); i++){
       if(!(isAlphaNumeric(p->getID()[i])) && !(p->getID()[i]=='_')){
        DEBUG_WM(DEBUG_ERROR,"[ERROR] parameter IDs can only contain alpha numeric chars");
        return false;
       }
    }
  }*/
```

```
      /*if (_params[i] == NULL || _params[i]->_length == 0) {
        DEBUG_WM(DEBUG_ERROR,"[ERROR] WiFiManagerParameter is out of scope");
        break;
      }*/
```

tablatronix:
I didnt even realize you could do this, I thought that was a constructor doh

PR #681, opened 2018-07-28
Fix custom html from out of scope

liebman:
Over eager "parameter out of scope" check:
Custom html was triggering "parameter out of scope" as its length was 0 causing parameters not to be added.  In my case I have a "label" (custom html) first and none of my 31 parameters were showing up.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -1115,7 +1115,7 @@ String WiFiManager::getParamOut(){
     char parLength[5];
     // add the extra parameters to the form
     for (int i = 0; i < _paramsCount; i++) {
-      if (_params[i] == NULL || (_params[i]->_length == 0 && _params[i]->_customHTML == NULL)) {
+      if (_params[i] == NULL || (_params[i]->_length == 0 && (_params[i]->_customHTML == NULL || _params[i]->_customHTML[0] == '\0'))) {
         DEBUG_WM(DEBUG_ERROR,"[ERROR] WiFiManagerParameter is out of scope");
         break;
       }
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -1115,7 +1115,7 @@ String WiFiManager::getParamOut(){
     char parLength[5];
     // add the extra parameters to the form
     for (int i = 0; i < _paramsCount; i++) {
-      if (_params[i] == NULL || _params[i]->_length == 0) {
+      if (_params[i] == NULL || (_params[i]->_length == 0 && _params[i]->_customHTML == NULL)) {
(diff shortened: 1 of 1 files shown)


liebman:
I think this should fix #678

tablatronix:
Thanks I was just about to look at this. I did not realize people were using custom html like this, its not reallly ideal.

liebman:
Yea, with 31 parameters I needed something to divide it into sections.

tablatronix:
So using a empty param and injecting custom html to insert arbitrary html sections?

Ill add an example for this so I can make sure I dont break it. Then maybe next release we can deprecate it and add some better way to insert html components. Or maybe each param can be a component with a callout or something, this would reduce memory on the stack.

liebman:
Basically this:
```
WiFiManagerParameter* section1 = new WiFiManagerParameter("<p>OTA Update</p>");
wm.addParameter(section1);
```

But I've put togeather a wrapper with callbacks for changed values and I add them all to a vector  to make it easier to manage its lifecycle:

```
    snprintf_P(message, sizeof(message), PSTR("<p>ESP: %s ATTiny: %u</p>"), SYNCHRO_CLOCK_VERSION, version);
    params.push_back(std::make_shared<ConfigParam>(wifi, message));

    params.push_back(std::make_shared<ConfigParam>(wifi, "position", "Clock Position", pos_str, 10, [](const char* result)
    {
        if (strlen(result))
        {
            uint16_t position = TimeUtils::parsePosition(result);
            dlog.info(FPSTR(TAG), F("setting position to %d"), position);
            if (clk.writePosition(position))
            {
                dlog.error(FPSTR(TAG), F("failed to set initial position!"));
            }
        }
    }));
    params.push_back(std::make_shared<ConfigParam>(wifi, "ntp_server", "NTP Server", config.ntp_server, 32, [](const char* result)
    {
            strncpy(config.ntp_server, result, sizeof(config.ntp_server) - 1);
    }));

    params.push_back(std::make_shared<ConfigParam>(wifi, "<p>OTA Update</p>"));
    params.push_back(std::make_shared<ConfigParam>(wifi, "ota_url", "OTA URL", getOTAurl(), 127, [](const char* result)
    {
        setOTAurl(result);
    }));
    params.push_back(std::make_shared<ConfigParam>(wifi, "ota_fp", "OTA fingerprint", getOTAfp(), 127, [](const char* result)
    {
        setOTAfp(result);
    }));
```

I'd love for this to be all handled internally with an interface like:
```
    wm.addParameter("ota_url", "OTA URL", getOTAurl(), 127, [](const char* result)
    {
        setOTAurl(result);
    }));
```
or something similar where there are callbacks for value changes or optionally the callback is called when when its not changed.

tablatronix:
I believe I fixed this, this PR can probably be closed.

PR #683, opened 2018-08-03
Hotfixes

ldab:
Add
```
WiFiManagerParameter::WiFiManagerParameter(const char *id, int length, const char *custom) {
  _id = id;
  _placeholder = NULL;
  _length = length;
  _value = NULL;

  _customHTML = custom;
}
```
This way [#L522](https://github.com/tzapu/WiFiManager/blob/32495668d8ac691d72ec77b5e423c6f8d3d845b7/WiFiManager.cpp#L522) to:
```
if (_params[i]->getValue() != NULL) {
```
Would allow to use custom HTML as <select> and drop-down list for example.

Because originally 
```
_id = NULL;
``` 
this line [[#L598]](https://github.com/tzapu/WiFiManager/blob/32495668d8ac691d72ec77b5e423c6f8d3d845b7/WiFiManager.cpp#L598) won't retrieve value from the drop-down list.

[screenshot_20180803-230205_chrome](https://user-images.githubusercontent.com/34199302/43639883-db1e6d0a-9771-11e8-8c96-13a5bbcca5e9.jpg)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+## Current development going on here :arrow_right: [Development Branch](https://github.com/tzapu/WiFiManager/tree/development)
+
 # WiFiManager
 ESP8266 WiFi Connection manager with fallback web configuration portal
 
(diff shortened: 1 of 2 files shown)


tablatronix:
Could you just add a none selected option ?

tablatronix:
```<option selected hidden>Choose here</option>```

PR #734, opened 2018-09-16
add callback before save config wifi or params

irwanayeah:
hi @tzapu @tablatronix previously I can't found listener before save wifi or params. i have made for it and very helpful. I use for disable other process can make stuck or reset ESP, enabled again when finish process.. and finally smooth process without ESP reset.

thx


tablatronix:
#452

PR #743, opened 2018-09-24
Set value

dontsovcmc:
Firstly, I correct your error: 
`if(_length < length){` to `if(_length < deflength){`
but it is dangerous to use strlen for default value cause overflow.
I remove you code to increate length. 

Secondly,
a. memset
b. need not _value[_length] = '\0'; case already memset with length+1

1. Now init() and getValue() are symmetrical for parameter length: you set and get with same limit.
2. init() parameter is safe: you can use EEPROM buffer without null terminator.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -71,7 +71,7 @@ void WiFiManagerParameter::setValue(const char *defaultValue, int length) {
   int deflength = strlen(defaultValue); // length actual
   // use the defult length if it's longer, 
   // @todo consider it might be useful to fail, so user knows they were wrong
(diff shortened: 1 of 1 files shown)


tablatronix:
geez how the hell did I do that.. thanks lol
yeah memset is probably ideal here

tablatronix:
I think you are correct that the length should not change if defvalue is longer, but you removed the checking, I would like to keep that in to throw debug errors. That is why I had the comment in there.

dontsovcmc:
sorry, i mean dangerous to allocate huge strlen return value

PR #687, opened 2018-08-05
issuu #644 add utf-8 charset to head

dontsovcmc:
https://github.com/tzapu/WiFiManager/issues/644

PR #772, opened 2018-11-26
separate charset UTF-8 and viewport meta

dontsovcmc:
https://github.com/tzapu/WiFiManager/pull/687


tablatronix:
applied in dev, will hotfix

PR #775, opened 2018-11-27
make `connectWifi` public method

TRANTANKHOA:
I'm using some default ssid and password that I want to try first before triggering the auto-connect workflow.

### DIFF
--- a/WiFiManager.h
+++ b/WiFiManager.h
@@ -74,6 +74,8 @@ class WiFiManager
     WiFiManager();
     ~WiFiManager();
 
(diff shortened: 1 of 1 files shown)


tablatronix:
Just use begin and check wifi status or waitforconnectresult, i would stick with esp functions for that

TRANTANKHOA:
But this built in function can set static IP too. Is there a consequence of making it public?

tablatronix:
It is not a public method

tablatronix:
you can do autoconnect with configportal disabled

PR #797, opened 2019-01-01
Added input field to enter a password for securing the device itself

sidey79:
Added basic authentication for ota update handler.

I've added an input fielf for a password, but i can't figure out where to save the variable.

image

I've tried to understand how static ip is saved and i think static ip is not preserved after reboot.

I need some help, where to go.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -962,6 +962,8 @@ void WiFiManager::handleWifi(boolean scan) {
   pitem.replace(FPSTR(T_v), WiFi_SSID());
   page += pitem;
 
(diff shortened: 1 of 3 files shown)


tablatronix:
Maybe user can pass in their own struct and deal with themselves?

sidey79:
@tzapu 

What do you think. Should be apply all fields via `WiFiManagerParameter ` class and then use the already implemented saveConfig behaviour ?

tablatronix:
With littlefs now in esp lib, this might be a little easier now and I am more open to adding persistent storage to wm, at least via a subclass or something

PR #803, opened 2019-01-08
wifimanager.cpp / wifimanager.h

sidey79:
I figured out, that there is no check on the hardware plattform, which is currently selected.
There are serval plattforms, which does not compile, i suggest adding a if defined check to the preprocessor.

Why is this useful?

I've a multi plattform Project. I used two seprerate source code projects in the past, but now i want to merge them. When i try to compile my project with an AVR Board, it fails, because it trys to compile the wifi lib but doesn't find serval used libs.

-added define that checks on esp32 or esp8266. If another hardware is selectec, then do not try to compile this code.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -12,6 +12,8 @@
 
 #include "WiFiManager.h"
 
+#if defined(ESP8266) || defined(ESP32)
+
 #ifdef ESP32
 uint8_t WiFiManager::_lastconxresulttmp = WL_IDLE_STATUS;
 #endif
@@ -2700,3 +2702,5 @@ void WiFiManager::WiFi_autoReconnect(){
     // }
   #endif
 }
+
+#endif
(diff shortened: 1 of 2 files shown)


tablatronix:
Do you thing we should include a #warning also just so it doesnt compile blind, I mean it is an esp exclusive library

sidey79:
Yea, it's an ESP lib, but my compiler isn't aware of what is an ESP lib, it will try to compile .h and .cpp files which are available. The reason is, i use project specific libs and not global arduino libs.


The warning may be an option, but users must enable verbose mode to see those warnings, so this will most people not help.


I think the travis check isn't caused by me.

sidey79:
What can i do to get this integrated?

tablatronix:
I will do it later

tablatronix:
Ugh Ill merge tomorrow, it wont let me since travis is failing atm.

Issue #367, opened 2017-05-08
WPS WorkaRound


Melcos1970:
I have made a small change to WiFiManager.cpp and .h to get a mid-term solution to the problem with WPS.
It's not the best way, but now you can use WPS if you are going to ..
I have no idea how to add suggestions to github, but I want to post them here so can add them if you want to use them.
The solution is actually simple .. I just add a button extra that makes a WPS call.

Changed:
Const char HTTP_PORTAL_OPTIONS [] PROGMEM = "<form action = \" / wifi \ "method = \" get \ "> <button> Configure WiFi </ button> </ form> <br/> <form action = \" / 0wifi \ "method = \" get \ "> <button> Configure WiFi (No Scan) </ button> </ form> <br/> <form action = \" / i \ "method = \" get \ "> <Button> Info </ button> </ form> <br/> <form action = \ "/ r \" method = \ "post \"> <button> Reset </ button> </ form> <br/> <Form action = \ "/ w \" method = \ "post \"> <button> WPS Setup </ button> </ form> ";

Inserted: lines +++
    Void handleReset ();
+++ void handleWPS ();
    Void handleNotFound ();

  Server-> on ("/ r", std :: bind (& WiFiManager :: handleReset, this));
+++ server-> on ("/ w", std :: bind (& WiFiManager :: handleWPS, this));
  // server-> on ("/ generate_204", std :: bind (& WiFiManager :: handle204, this)); // Android / Chrome OS captive portal check.
  Server-> on ("/ fwlink", std :: bind (& WiFiManager: handleRoot, this)); // Microsoft captive portal. Maybe not needed. Might be handled by notFound trades.

Inserted section / ++ ++ / after void WiFiManager :: handleReset () section
/ ++
Void WiFiManager :: handleWPS () {
  DEBUG_WM (F ("WPS Start"));

  String page = FPSTR (HTTP_HEAD);
  Page.replace ("{v}", "Info");
  Page + = FPSTR (HTTP_SCRIPT);
  Page + = FPSTR (HTTP_STYLE);
  Page + = _customHeadElement;
  Page + = FPSTR (HTTP_HEAD_END);
  Page + = F ("Module will start WPS in a few seconds.");
  Page + = FPSTR (HTTP_END);
  Server-> send (200, "text / html", page);

  DEBUG_WM (F ("Sent WPS page"));
  // WPS works only in STA (Station mode) only.
  WiFi.mode (WIFI_STA);
  delay (1000);
[...]

PR #827, opened 2019-02-17
WPS Push Button

peastone:
This pull request addresses https://github.com/tzapu/WiFiManager/issues/367
This is how a solution might look like.

I know you probably will not merge against master, but I had the code already adapted.
Before I will delete it, it may serve as inspiration.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -146,6 +146,7 @@ void WiFiManager::setupConfigPortal() {
   server->on(String(F("/")), std::bind(&WiFiManager::handleRoot, this));
   server->on(String(F("/wifi")), std::bind(&WiFiManager::handleWifi, this, true));
   server->on(String(F("/0wifi")), std::bind(&WiFiManager::handleWifi, this, false));
+  server->on(String(F("/wps_pbc")), std::bind(&WiFiManager::handleWpsPushButton, this));
   server->on(String(F("/wifisave")), std::bind(&WiFiManager::handleWifiSave, this));
   server->on(String(F("/i")), std::bind(&WiFiManager::handleInfo, this));
   server->on(String(F("/r")), std::bind(&WiFiManager::handleReset, this));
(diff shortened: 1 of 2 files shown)


tablatronix:
Have you tested this ? I thought WPS was broken..

peastone:
I tested this and it works on my device (Sonoff S20, ES8266).

```
root@LEDE:~# hostapd_cli -i wlan0 wps_pbc
OK
root@LEDE:~# ping ESP_134ED9
PING ESP_134ED9 (192.168.254.142): 56 data bytes
64 bytes from 192.168.254.142: seq=6 ttl=255 time=7.829 ms
64 bytes from 192.168.254.142: seq=7 ttl=255 time=39.306 ms
64 bytes from 192.168.254.142: seq=8 ttl=255 time=65.007 ms
```
Here, you see that the ESP connects as it responds to pings again.

I also found out while testing, that it is very important to switch back to AP mode after WPS.
If WPS was always successful, that would not be a problem.
But if WPS fails for some reasons (... too little time to press the button on the router ...), than you will hang in station mode having no connection.

What's interesting though, is that the WPS fails very quickly, if you do not reach the router within 10 to 15 seconds. I just did one try, where I did not press the WPS button on the router. I think, I measured something around 13 seconds. After that time, the ESP access point showed up again, means it knew it failed after that time.

tablatronix:
Maybe we can add a js delay!

peastone:
I suggest a delay of 30 seconds before the ESP will start with WPS. If you do not manage to reach the button of your router in that time, you can try again. I would not choose more than 30 seconds. That might be too long.

By the way, if you like the change, will you put a new release to platformio? I'd like to use the fix in my project. WPS is a killer feature :wink:.

tablatronix:
Next release is development branch, you would have to pr it against that

peastone:
I don't want to PR against development branch, as long as it fails to build: https://travis-ci.org/tzapu/WiFiManager/builds/493361281

I cannot reproduce quite the same issue as in the build. I have Arduino 1.8.8 installed and board package esp8266 version 2.4.0-rc2. There, I get the message:

> Error: led: Invalid option for board "generic"

The same applies for "LwIPVariant", "DebugLevel".

So, do you know, why these options are set? Otherwise, can we remove them?

tablatronix:
Build fails because the examples are being rewritten and have to be fixed.

tablatronix:
Ohhh you mean travis script, umm that was added at some point to get something to pass, there was issues with lwip  can maybe be removed now, wonder why it complains for generic esp8266, i will follow up and get this fixed, #523

tablatronix:
Can you refactor this against development, or a new feature branch off of it

josejsarmento:
I've been trying to implement this on the most recent commit on main with no success, does anybody have a working version?

PR #894, opened 2019-06-02
signal class value has been wrapped to int()

mikdatdogru:
i wrapped to int() function, because cannot match on css because the incoming data is decimal

image


tablatronix:
Did map() always return a float?

mikdatdogru:
yeap round() function returns float.

PR #895, opened 2019-06-02
Update WiFiManager.template.html

mikdatdogru:
parse.js parsing error fixed


tablatronix:
I do not really use parse anymore on this , I had plans to rework that but I have no idea if it still works or is valid, I mostly just added stuff to it for testing.

mikdatdogru:
yes I noticed the situation. in the master branch I was using the parsed data directly. but developer branch is different :)

tablatronix:
Yeah it uses a ton of strings so it would require a little rework

PR #928, opened 2019-08-30
hotfix: rename internal HTTP_HEAD to HTTP_HEADER 

d-a-v:
ref: https://github.com/esp8266/Arduino/pull/6413

### DIFF
--- a/WiFiManager.h
+++ b/WiFiManager.h
@@ -22,10 +22,10 @@ extern "C" {
   #include "user_interface.h"
 }
 
-const char HTTP_HEAD[] PROGMEM            = "<!DOCTYPE html><html lang=\"en\"><head><meta charset=\"UTF-8\" name=\"viewport\" content=\"width=device-width, initial-scale=1, user-scalable=no\"/><title>{v}</title>";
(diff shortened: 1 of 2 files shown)


WiFiManager.h
```diff
@@ -22,10 +22,10 @@ extern "C" {
   #include "user_interface.h"
 }
 
-const char HTTP_HEAD[] PROGMEM            = "<!DOCTYPE html><html lang=\"en\"><head><meta charset=\"UTF-8\" name=\"viewport\" content=\"width=device-width, initial-scale=1, user-scalable=no\"/><title>{v}</title>";
+const char HTTP_HEADER[] PROGMEM            = "<!DOCTYPE html><html lang=\"en\"><head><meta charset=\"UTF-8\" name=\"viewport\" content=\"width=device-width, initial-scale=1, user-scalable=no\"/><title>{v}</title>";
```
marcelstoer: That `HTTP_` is wrong/misleading/imprecise anyhow IMO. It should be `HTML_`.

tablatronix: Yes, but then they all have to change also.. Ill fix it in development branch its a major ver bump

PR #984, opened 2019-12-22
Add standalone AP support and merge redundant options.

Petapton:
-added option to create standalone Access Points (AP) instead of joining an existing one.

-merged options 'Configure WiFi' and 'Configure WiFi (no scan)' into 
'Join network'. User can still configure custom wify entry, just 
ignoring scanned ones and inserting SSID and password into respective 
fields.


tihoangyeudau:
It is usefull. Hope it works

tablatronix:
I don't get it.. Those are not redundant they are 2 separate menu options, one skips scanning..

Also why create aps? You can just add a parameter and then use that in your own code to run your own aps if you wanted..

tihoangyeudau:
because there is no option for ap or connect to wifi router in wifimanager interface

tablatronix:
What do you mean, provide and example of why you would want this?

Petapton:
> I don't get it.. Those are not redundant they are 2 separate menu options, one skips scanning..
> 
> Also why create aps? You can just add a parameter and then use that in your own code to run your own aps if you wanted..

Yes, but one is just a submenu of the other one. You could just delete the no scan option since it has no advantages over the scan one.

Petapton:
> I don't get it.. Those are not redundant they are 2 separate menu options, one skips scanning..
> 
> Also why create aps? You can just add a parameter and then use that in your own code to run your own aps if you wanted..

It's far less straightforward and you have to implement it on your own. It's also less clean, 'cause you don't have a dedicated section, you just place AP settings underneath other settings that are more or less unrelated with them, making the whole system more confused and confusing ('Why the hell I wanted to connect to a hotspot and you also prompt me to create my own??').
I think not everyone thinks of such a solution. I, for example, didn't. :D

tablatronix:
Its a good idea, but not how I would implement it.
I will create a new issue for this , also people have asked for custom ap name in settings also.

PR #945, opened 2019-09-20
Two examples ArduinoJson 6 compatibility changes

bigjohnson:
Changed two examples code for ArduinoJson 6 library usage.

PR #812, opened 2019-01-21
Update AutoConnectWithFSParameters to new ArduinoJson library

jeroenrnl:
Fixes "DynamicJsonBuffer was not declared in this scope"

Works with ArduinoJson library 6.7.0beta (current version in library
manager)

Should fix #654

PR #993, opened 2020-01-14
Fix ArduinoJson (Using ARDUINOJSON_VERSION_MAJOR)

Ruandv:
This is relates to the ArduinoJson 5 / 6 changes.
I took what @bigjohnson did in a502456 and just added a #define UseJsonV6 option that can be commented out if you want to use ArduinoJsonV5.
Tested on ArduinoJson V 5.13 and ArduinoJson V 6.13 (Latest as at 14/01/2020)

This also links to #654 and #812 
- [x] update examples to be compatible with future arduinojson library
- [x] Maybe add support in the example in question for both versioons somehow?
- [x] Can also Close #945

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -19,13 +19,24 @@ https://github.com/esp8266/Arduino/blob/master/doc/Troubleshooting/debugging.rst
 ### Basic Infos
 
 #### Hardware
-WiFimanager Branch/Release:  Development
+WiFimanager Branch/Release:
+- [ ] Master
+- [ ] Development
 
-Esp8266/Esp32: 
+Esp8266/Esp32:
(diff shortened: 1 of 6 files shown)


bigjohnson:
Perfect, but I think is better automate it replacing
#ifdef UseJsonV6
with
#if ARDUINOJSON_VERSION_MAJOR >= 6

Ruandv:
Ok good suggestion. Thank you I didn't know there is a way to do that. Will update it soon

tablatronix:
I should have rebased this.. Should be ok

tablatronix:
Compiles for me

userweb2018:
For me too, but this is strange....
It is not working perfectly everytime!
In which way can I use it with ondemand with pushbuttom and parameters?

PR #1005, opened 2020-02-07
trivial: WiFiManagerParameter get accessors should work on const

flavio-fernandes:
Enhancement: It should be okay to invoke getID() and getValue() in a
const instance of the WiFiManagerParameter class.

For usage example, see: https://github.com/flavio-fernandes/persistWifiSetting/commit/135adb1c7e436db1ed92ccd5eb522b9ae418b216

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -88,25 +88,25 @@ void WiFiManagerParameter::setValue(const char *defaultValue, int length) {
     strncpy(_value, defaultValue, _length);
   }
 }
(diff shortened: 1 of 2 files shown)


tablatronix:
Could this break anything ?

flavio-fernandes:
> Could this break anything ?

I don't think so. If anything, it is just making it possible to be used by a constant instance of the class.
Since it returns cont char* and a int, there is no changes to the instance.
If it broke something we would see it at compile time, not run time.

PR #1011, opened 2020-02-18
Update WiFiManager.cpp

marcovannoord:
Fixed bug where, after clicking "Save", it would display an empty page to the user, instead of the wifi save page. (due to local variable)


tablatronix:
Thanks for checking on this

PR #1017, opened 2020-03-05
 Fix opening hotspot in autoconnect after reset

lu4p:
I noticed that a hotspot is opened if the ESP8266 reconnects to a previous wifi using autoconnect after a reset. 

I use the same workaround like in the [shutdownConfigPortal](https://github.com/tzapu/WiFiManager/blob/4a98ba0e29606c684e66fa80e53f1293a269540b/WiFiManager.cpp#L700) method.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -317,6 +317,20 @@ boolean WiFiManager::autoConnect(char const *apName, char const *apPassword) {
     DEBUG_WM(F("STA IP Address:"),WiFi.localIP());
     _lastconxresult = WL_CONNECTED;
 
+    DEBUG_WM(DEBUG_VERBOSE,F("disconnect configportal"));
+    bool ret = false;
+    ret = WiFi.softAPdisconnect(false);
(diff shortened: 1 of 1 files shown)


tablatronix:
huh?

tablatronix:
You realize the esp automatically starts the ap right? Just set your mode properly in your code and it will remember it..

lu4p:
I know that.
I had an issue on the wemos d1 mini where a WiFi with the SSID of 'ESP-id' is started, when autoconnect succeeded from previously saved credentials, it also connected to the configured WiFi.

My normal Configportal WiFi prefix was also another than ESP. And the autoconnect Configportal exited

This issue is fixed by this PR so that the WiFi Hotspot is not started on successful autoconnect.

tablatronix:
You do not understand how esp works.
set
WiFi.mode(WIFI_STA);
this has nothing to do with wm

lu4p:
OK sorry my bad thanks anyway.

tablatronix:
Did it work?

PR #1043, opened 2020-04-30
Pull request horendus

horendus:
I needed more callbacks options from the library to trigger some voice commands on my new project ("Device Connected" and "Device Disconnected")

setAPClientConnectedCallback
setAPClientDisconnectedCallback

While I was at it I also added (F to some of the debug messages to reduce the RAM usage of the library.


tablatronix:
cannot merge, 

adding to #452 for callback proposals

PR #1053, opened 2020-05-14
Readme: fill DNS server if setSTAStaticIPConfig

dontsovcmc:
I think it's better to add Note about DNS if somebody want to use static IP and syncronize time (NTP) or request web servers by hostname in source code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,6 +292,8 @@ wifiManager.setSTAStaticIPConfig(IPAddress(192,168,0,99), IPAddress(192,168,0,1)
 ```
 There are a couple of examples in the examples folder that show you how to set a static IP and even how to configure it through the web configuration portal.
 
+NOTE: You should fill DNS server if you have HTTP requests with hostnames or syncronize time (NTP). It's the same as gateway ip or a popular (Google DNS: 8.8.8.8).
+
 #### Custom HTML, CSS, Javascript
 There are various ways in which you can inject custom HTML, CSS or Javascript into the configuration portal.
 The options are:


tablatronix:
Could also add a debug if dns is empty, saying no dns, hostnames will not resolve or something.

Is it possible to get the dns server or assume it is gateway if unset?

dontsovcmc:
"Is it possible to get the dns server or assume it is gateway if unset?" 
https://www.quora.com/How-does-the-computer-know-which-DNS-server-to-use#
The main idea: routers has DNS-relay function, that proxies DNS requests to ISP DNS servers, obtained by router.
I think Debug message if DNS is empty needed. "Hidden magic" DNS=gateway isn't good practise.

tablatronix:
true but typically your dns server is going to be the same one if you were dhcp.. so can we query that server and cache it etc. Does dhcp provision dns servers?

dontsovcmc:
It's bad idea to make hidden requests to dhcp server. One of the reason of using static ip when you don't have dhcp server.

dontsovcmc:
Here is the same problem: https://github.com/esp8266/Arduino/issues/2371#issuecomment-536366080

chupocro:
> "Is it possible to get the dns server or assume it is gateway if unset?"
> https://www.quora.com/How-does-the-computer-know-which-DNS-server-to-use#
> The main idea: routers has DNS-relay function, that proxies DNS requests to ISP DNS servers, obtained by router.
> I think Debug message if DNS is empty needed. "Hidden magic" DNS=gateway isn't good practise.

A few years ago static IP worked without specifying DNS - probably the same IP as gateway was used as DNS IP if DNS wasn't specified in `wifi.config()`

I have some older code from 2016. and 2017. I had to rewrite for compiling it with present libraries because the devices couldn't connect to the server with the old code until I added DNS. Someone will probably confirm the IP used for DNS was the same as gateway in the old libraries.

PR #1054, opened 2020-05-14
notify message about dns

dontsovcmc:
here it is


tablatronix:
Hmm what does this mean? Not sure I understand

tablatronix:
Closing, not sure what this is trying to indicate to users

PR #1091, opened 2020-07-08
Adding dependencies to library.json to facilitate working with platformio

lazyzero:
I added the dependencies section to facilitate the work with platformio. If missing, I had to add DNSServer, WebServer, WiFi and FS manually to the lib_deps section of the platformio.ini of my ESP32 projects using WiFiManager.

Best Christian

### DIFF
--- a/library.json
+++ b/library.json
@@ -26,4 +26,21 @@
     "espressif32"
   ],
   "version": "2.0.3-alpha"
+  "dependencies":
+  [
+      {
+          "name": "DNSServer",
+          "version": "1.1.0"
+      },
+      {
+          "name": "WebServer",
+          "version": "1.0"
+      },
+      {
+          "name": "WiFi",
+          "version": "1.0"
+      }
+  ]
+
+
 }
--- a/library.json
+++ b/library.json
@@ -25,7 +25,7 @@
     "espressif8266",
     "espressif32"
   ],
-  "version": "2.0.3-alpha"
+  "version": "2.0.3-alpha",
   "dependencies":
   [
       {
@@ -38,6 +38,10 @@
       },
       {
           "name": "WiFi",
+          "version": "1.0"
+      },
+      {
+          "name": "FS",
           "version": "1.0"
       }
   ]


tablatronix:
hmm arent these all included in esp32 platform ?

lazyzero:
Hi @tablatronix 

the libraries mentioned are not core component, they are in the libraries folder of https://github.com/espressif/arduino-esp32 So I assume they are not included in the build process by default. 

Using a platformio.ini like this:

`
platform = espressif32
board = esp32doit-devkit-v1
framework = arduino

lib_deps =
  https://github.com/tzapu/WiFiManager.git#development
`
as espressif32 is used in version 1.12.2 and once tested againd 1.8.0 result in 

`
.pio/libdeps/esp32doit-devkit-v1/WifiManager/WiFiManager.h:62:22: fatal error: WiFi.h: No such file or directory
 
**************************************************************
* Looking for WiFi.h dependency? Check our library registry!
*
* CLI  > platformio lib search "header:WiFi.h"
* Web  > https://platformio.org/lib/search?query=header:WiFi.h
*
**************************************************************
 
compilation terminated.
`
replacing the lib_deps by:
`
lib_deps =
  https://github.com/tzapu/WiFiManager.git#development
  WiFi
  WebServer
  DNSServer
  FS
`

is fixing it. Adding the dependencies to library.json of WiFiManager has the same effect.

Hope this helps to reproduce the problem. Am I over-viewing something?

Best Christian

tablatronix:
maybe there is something wrong with your pio dependancy linker settings.. There are several options that can get changed, but afaik core libs for platform should always be in the path...

```
[env:esp32doit-devkit-v1]
platform = espressif32
board = esp32doit-devkit-v1
framework = arduino
```

works for me...
What does your compile output look like , use verbose..

tablatronix:
You know I have seen this wifi.h problem before, hmm I wonder if its a bug in pio, I saw it a few weeks ago.

lazyzero:
I think I have the difference in our pio settings. What is the output of this line for you?

`LDF Modes: Finder ~ chain, Compatibility ~ soft`

chain is the default value for the ldf. By changing it to deep, I can compile without any modifications.

lazyzero:
in chain mode also setting these headers for esp32 helps the ldf to find the dependencies.
#include <DNSServer.h>
#include <WiFi.h>
#include <WebServer.h>

But I think that is not the way to go. including headers that are not used in the file is not so nice and degrades the readability of the code.

tablatronix:
Since these are core libs I do not think the should be in the pio file, but someone else that knows might have better input

PR #1106, opened 2020-08-13
Fix warning in WiFiManager::debugPlatformInfo for ESP32

davefiddes:
This fixes an unused variable warning as the call to
heap_caps_get_free_size() is not required.

Tests:
 - Enable debug mode and verify free heap and ESP SDK
    version are correctly displayed
 - Cross check with the free heap reported via the captive
    portal info page


tablatronix:
Ah yeah that was from before they added the wrappers, thanks

Issue #417, opened 2017-09-08
bssid support


tablatronix:
Allow bssid saving

- [ ] work out use interface
- [ ] thoroughly test begin with bssid saving

Instead of adding a ton of options, maybe make bssid implicit when showdupaps is true ?

tablatronix:
#221

tzapu:
test cases for reconnect with begin()
- 2 aps with same ssid/pwd, different bssid, ssid, pwd & bssid given to begin
- 2 aps with same ssid/pwd, different bssid, ssid, pwd given to begin (does it store bssid for the connected network if not given?)
- 2 aps with same ssid/pwd, different bssid, ssid, pwd & bssid given to begin, the connected bssid leaves (does it hang or reconnect to the other one even if bssid is different)

i was under the impression that bssid was not saved, but i did not check it in a looooong while

tablatronix:
I am pretty sure it works, i tested it I think, and i am in the process up getting my openwrt array up( its just 2 devices) But it will let me test ap enviroments better.

flexifi:
Does BSSID is now implemented

tablatronix:
No

flexifi:
you mean BSSID is now working sir

tablatronix:
Not in wifimanager, this issue is still open

PR #1114, opened 2020-08-22
Add Fast connect support

roberttidey:
The optional fast connect support uses the version of WiFi.begin with channel and router mac address defined. This avoids scanning operations during normal start up and reduces the start up connect time considerably (from 3 seconds down to 1 second).

This is particularly beneficial in deep sleep applications where the faster start up makes it more responsive to a wake and minimises power usage if on battery.

Default functionality is left as before. Fast connect is activated using the setFastConnectMode(true) call. The channel and mac address are retrieved automatically when required.

Changes have been made in development branch

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -821,7 +821,12 @@ bool WiFiManager::wifiConnectNew(String ssid, String pass){
   DEBUG_WM(DEBUG_DEV,F("Using Password:"),pass);
   WiFi_enableSTA(true,storeSTAmode); // storeSTAmode will also toggle STA on in default opmode (persistent) if true (default)
   WiFi.persistent(true);
-  ret = WiFi.begin(ssid.c_str(), pass.c_str());
+  if(_fastConnectMode) {
+    getFastConfig(ssid);
+    ret = WiFi.begin(ssid.c_str(), pass.c_str(), _fastConnectChannel, _fastConnectBSSID, true);
+  } else {
+    ret = WiFi.begin(ssid.c_str(), pass.c_str());
+  }
   WiFi.persistent(false);
   if(!ret) DEBUG_WM(DEBUG_ERROR,"[ERROR] wifi begin failed");
   return ret;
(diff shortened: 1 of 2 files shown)


tablatronix:
wifiConnectNew could check to see if we already have a scanresult , instead of scanning.
wifiConnectNew could avoid scanning by connecting once, then getting the bssid and channel, and setting it after.

These would speed up new connection saving, also I intend to eventually add this capability to the webinterface as toggles

* remember channel
* remember bssid

tablatronix:
Also add documentation 

* On new connection, use esp capability to store along with ssid and pass, the channel and BSSID of connection
* this enables faster connection because it uses the bssid instead of the ap name to connect to your ap
* if connection fails, perhaps due to channel change, attempts (1) omitting fastconnect
* fast connect variables are stored in flash by the ESP library automatically using native persistent methods

* test ESP32 for any required variety

roberttidey:
wifiConnectNew is only called if the normal connect mode without parameters fails. Once the scan has found the right network then the channel and BSSID are stored by the underlying esp framework and re-used automatically from then on. Only if channel of BSSID changes does a new scan get invoked so it has zero impact on regular connection times

I didn't think there was a reliable mechanism for getting the BSSID / channel without performing a scan. Retrieving these after a connection just returns previously stored parameters not the actual values of the active connection.

tablatronix:
Ahh, good point it might just be the config values.

Wificonnectnew is called when saving, scans can be slow and most likely we already have a cached one so it is something I might look at.

I also have to update this and fix any conflicts

TheNetStriker:
I have a question about this new feature. At my home I have two access points with the same SSID and I had the problem that my ESP32 connected randomly to one of those access points. One of the access points is much further away than the other and when my ESP32 connected to it the connection was really bad. With this new feature the devices now always connect to the access point with the best RSSI which is much better.

The only problem that remains is that when I e.g. update the firmware of my access points the devices will eventually connect to the "bad" access point again because the other one has not restarted yet. Would it be possible to implement something to force the devices to only connect to a specific BSSID?

tablatronix:
Hmm well you can force bssid and we could add no fallback options

TheNetStriker:
@tablatronix Thanks for the hint. I solved this problem by simply storing the current BSSID in a variable after connecting. For the reconnect I've built my own method that takes the SSID and password from the WifiManager and my stored BSSID to connect to the wifi again. The ESP then waits until the access point that was connected at startup is online again and does not connect to the other access point. Here is how I've done it:

```
uint8_t bssid[6]; // Global variable to store BSSID
bool success = wifiManager.autoConnect(hostname);
// Store current BSSID and channel for later reconnect
memcpy((void *) &bssid[0], (void *) WiFi.BSSID(), 6);
```

Then in my reconnect method I simply use this line to reconnect:

`WiFi.begin(wifiManager.getWiFiSSID().c_str(), wifiManager.getWiFiPass().c_str(), 0, bssid, true);`

But it would be nice to be able to force a BSSID at startup in the wifiManager.autoConnect method.

tablatronix:
#417

TheNetStriker:
What is the status of this feature? Will this be merged into the master branch anytime soon?

tablatronix:
There has been no development on this, but bssid and channel storage should be easy to add now

roberttidey:
I use the fast_connect branch all the time now and it works well with the automatic detection of bssid and channel giving the fast connect.

It would be good to have it or equivalent in main branch.

tablatronix:
really? Lol I forgot about this, Ill take a look soon

tablatronix:
see issue

PR #1140, opened 2020-11-02
Callback

crocsg:
Hello,
i was working on a project where i needed to blink some led while wifi connection and when ap portal is active and need to be configurate. So i added 2 callbacks in WifiManager class

Regards


tablatronix:
You can use non blocking mode in the alpha ver

PR #1151, opened 2020-11-22
Use constrain properly

lbussy:
As written, the way constrain is used has no effect/unintended effect and will generate the following warnings:

```
warning: second operand of conditional expression has no effect [-Wunused-value]
warning: third operand of conditional expression has no effect [-Wunused-value]
```

This change addresses those warnings and corrects the behavior.


tablatronix:
Thanks, no idea why I thought it was edit by reference, been using php too much lately.. lol

lbussy:
I stared at it for a while wondering what was wrong.  Looked legit to me as well.

PR #1168, opened 2020-12-12
use 6 FSParameters and edit with a template JSON

ligantx:
I found adding FSParameters more complicating than just crate a JSON at the beginning with all the information needed. Also, i added more parameters, so now there are 6 easy to edit parameters. Just edit the commented JSON using arduinojson.org/v6/assistant/ and paste the result.

### DIFF
--- a/examples/AutoConnectWith6JsonFSParameters/AutoConnectWith6JsonFSParameters.ino
+++ b/examples/AutoConnectWith6JsonFSParameters/AutoConnectWith6JsonFSParameters.ino
@@ -1,0 +1,299 @@
+#include <FS.h>                   //this needs to be first, or it all crashes and burns...
+#include <ESP8266WiFi.h>          //https://github.com/esp8266/Arduino
+//needed for library
+#include <DNSServer.h>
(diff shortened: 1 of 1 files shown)


tablatronix:
Yeah I always hated that, on my list is to find a better way to programatically make params. I was going to wait until replcing spiffs with littelfs and then working on a good example or wrapper class ( you can not make child classes of wm param )

ligantx:
it can be made as loop, but its not so important as its just 6 parameters..
Also, it cannot be really generic for unlimited parameters as the Json size should be changed by hand (using ArduinoJSON assistant)
What i dont really understand from WiFiManager examples, is why we dont just search for config.json file. If it does not exist THEN do all the parameters init, mappings, wifimanager initializations etc.
If exists, then skip everything and use your config file for the rest of the code..

tablatronix:
WM does not depend on any FS

ligantx:
That makes sense. Because all the platforms i worked, supported FS, so i was curious. 
---
Did you test my code?

PR #1175, opened 2020-12-16
Fix. The library is not assembled under ESP8266

egrekov:
Fix [commit](https://github.com/tzapu/WiFiManager/commit/e071e3ad8fb799044589643b1f355ef750485599). The library is not assembled under esp8266.


tablatronix:
Thanks, I saw this late last night, and thought I fixed it

tablatronix:
I was testing some stuff, for some reason my esp32 is so unstable I can barely use it, sorry about that

PR #1182, opened 2020-12-31
Support callbacks when waiting on WiFi connection and when in AP mode

plutino:
This PR adds 2 configurable callbacks to allow status light update.  E.g., quick flashing during WiFi connection and slow flashing when the board is in AP mode waiting for WiFi configurations.  

Two public methods are added:

1. `setConnectionWaitCallback()`.
This function sets a callback that is invoked periodically when waiting for a WiFi connection.  Two changes are made to support this:
    1. Don't use the blocking `WiFi.waitForConnectResult()`.  Instead, setting the default connection timeout to 60 seconds, which is the same as the default timeout for `WiFi.waitForConnectResult()`.
    2. Change all blocking `delay` calls to non-blocking `while` loop and time check.  This allows the callback to be used for pin update.

2. `setConfigWaitCallback()`.
This function sets a callback that is invoked periodically when the board is in AP mode waiting for a new WiFi configuration.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -349,9 +349,6 @@ int WiFiManager::connectWifi(String ssid, String pass) {
 }
 
 uint8_t WiFiManager::waitForConnectResult() {
-  if (_connectTimeout == 0) {
-    return WiFi.waitForConnectResult();
-  } else {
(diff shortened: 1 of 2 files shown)


tablatronix:
Deprecated by development non blocking mode

PR #1195, opened 2021-01-25
Not saving custom IP and parameters in a certain case.

EgHubs:
There is a case that I encountered that causes the ESP parameters and custom IP not to be saved, which happens when the user enters a new WiFi data and parameters at the configuration point and the ESP fails to connect to that WiFi network -which could possibly be a problem related to that router network- so, in that case, I think we should save the WiFi custom IP and user parameters anyway.


EgHubs:
Sorry this was my first PR so I couldn't do it the right way, so I will close it and reopen a new one

tablatronix:
use shouldbreakonconfig() and handle as you normally would

or param callbacks in 2.x

PR #1199, opened 2021-01-28
adding WifiManager::isConfigPortalActive() abd WifiManager::isWebPortalActive()

cpainchaud:
because sometimes you need to know :p !!!!!

### DIFF
--- a/WiFiManager.h
+++ b/WiFiManager.h
@@ -640,6 +640,10 @@ class WiFiManager
       // DEBUG_WM("NO fromString METHOD ON IPAddress, you need ESP8266 core 2.1.0 or newer for Custom IP configuration to work.");
       return false;
     }
+	
+  public:
+    inline bool isConfigPortalActive() { return configPortalActive; }
+    inline bool isWebPortalActive() { return webPortalActive; }
 };
 
 #endif


tablatronix:
Makes sense I assume users would do this in their own code.

tablatronix:
changed method names

```C++
    // check if config portal is active (true)
    bool          getConfigPortalActive();
    
    // check if web portal is active (true)
    bool          getWebPortalActive();
```

cpainchaud:
@tablatronix  about the inline thing : I did it for performance and also final image size. It would be such a waste to do a CALL for this!
What do you think ?

tablatronix:
Does compiler not optimize this out anyway?

tablatronix:
my code is 16 bytes smaller, I have not checked the asm.

cpainchaud:
It's not always I see in the field so I usually optimize what I can by myself :D

cpainchaud:
@tablatronix  Here is the generated ASM output without inline:

.LVL26:
	.loc 1 166 0
	l32r	a2, .LC21	#, tmp48
	mov.n	a10, a2	#, tmp48
	call8	_ZN11WiFiManager20isConfigPortalActiveEv	#

Which does not exist when I am using INLINE

tablatronix:
Thanks ill look at it, I would love to move on to optimizing everything now for size and speed not that its fairly stable.

my dumbass was mistakingly testing without using the actual funcs, so they were compiled out lol geez

Testato:
> @tablatronix about the inline thing : I did it for performance and also final image size. It would be such a waste to do a CALL for this!
> What do you think ?
Performance or image size, it's not possible have the two togheter.
Inline duplicate the function code every time you use it, so the image size is increased.
Normally on embedded system the image size is preferred so inline is not used.
This two functions, imho, are not a critical speed function

cpainchaud:
Hi!

 My tests are showing image is small with inline, which makes sense because 1/ you have  some bytes for the function itself and second some instructions to prepare each call

cpainchaud:
@Testato 

Without inline image size is 971584 
With inline it is 971576

tablatronix:
Thanks, all, if you want feel free to hit up the gitter to discuss in depth

PR #1204, opened 2021-02-04
fix  debug message in WiFiManager::handleParamSave

cpainchaud:
DEBUG_WM(DEBUG_VERBOSE,F("<- HTTP Wifi save "));  while it's obviously Params instead of Wifi


tablatronix:
Thanks I saw this the other day, and forgot to fix it

PR #1206, opened 2021-02-04
fix potential memory leak in  WiFiManagerParameter::setValue

cpainchaud:


tablatronix:
So this would address a memory leak when calling setValue multiple times with changing values, correct?

cpainchaud:
I think so

tablatronix:
I need reviewers lol, ill try to look at this when i get a chance

tablatronix:
reverted, crashes esp32

bad array access

assertion "heap != NULL && "free() target pointer is outside heap areas"

tablatronix:
@cpainchaud I had to revert this it was crashing esp32

cpainchaud:
I saw that, it's weird actually, it means that another part of the code is relying on that memory to never be freed while it's obviously leaked ....

tablatronix:
yeah I did not really look at it at all, was busy on stuff. it crashed immediately also which is odd..

hurricanefrog:
This is because _value is not initialized in the constructor, or rather the `init` function - meaning the pointer is random.

Simply do `_value = nullptr;` in `init` and it's fixed

tablatronix:
I think this is right, please check

hurricanefrog:
Works fine, doesn't crash

hurricanefrog:
> Works fine, doesn't crash

Or maybe it does? Just re-tested this. Initial flash goes fine, but if I OTA update "my" application again, I get a `StoreProhibited` error. Need to check where exactly, though

hurricanefrog:
I think this is because `_length` is uninitialized as well! So `if(_length != length)` yields random results

tablatronix:
Thanks for testing this, let me know what you find or if you do not, etc

zenz:
on ESP8266, this
delete[] _value;
_value = new char[....]
will cause a crash.
So I use unique_ptr to try to solve the issue.
And I make a pull request, please check.

tablatronix:
Thanks, I havent had time to really look at this, and I am pretty sure I left a mic of null and nullptr in here, It need to be reworked to be memory safe, and I was just trying to figure out how make this safer when these are out of scope ( which alot of people do ) and make it not crash

PR #1244, opened 2021-04-23
Bad portal settings callback

flynny75:
Add a new callback to be invoked when portal settings didn't work


tablatronix:
Can you provide a use case or example ?

flynny75:
I have an orange and green status LED.

I am using setAPCallback to light the orange when the device needs to be configured via the portal, setSaveParamsCallback to start the green LED flashing to indicate that the settings the user has entered have been accepted and something is happening, and the new setNewSettingsFailed to stop the green flashing and turn the green LED off to indicate that the newly entered settings didnt work, and the user should return to the portal to update the settings.

PR #1258, opened 2021-06-09
Added Github Action to Compile Examples on ESP8266

shafr:
Hi, I've added compilation of all examples, so you would see if something would fail here in Github.
There are few tests failing - not sure what version of ArduinoJson it should be using - latest 6 fails with missing methods.

Please squash while merging.

### DIFF
--- a/.github/workflows/examples_esp8266.yml
+++ b/.github/workflows/examples_esp8266.yml
@@ -54,7 +54,7 @@ jobs:
         https://github.com/tzapu/WiFiManager \
         https://github.com/bblanchon/ArduinoJson \
         https://github.com/knolleary/pubsubclient \
-        https://github.com/knolleary/sstaub/Ticker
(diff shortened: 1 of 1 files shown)


tablatronix:
Yeah there is an example that needs to be updated I think I just need to copy it over from hotfixes or a pr

tablatronix:
Thanks , I will try to clean up examples, need to add esp32

saikek:
Will add ESP32, ok

tablatronix:
Thanks all , travis is down (migrate to .com required) I do not have repo access to migrate it, so this will be useful. I will fix the arduinojson example and just remove 5.x entirelty.

tablatronix:
- add pull_request ?
- add branches filter (master)

to trigger

tablatronix:
Thanks I will look it over and rework as needed, I will be cleaning up examples anyway for a beta release soon

PR #1264, opened 2021-06-22
CI / DC Update - Library Build, Linting, Examples Json Fix

shafr:
**Updated:**
```
Examples compilation in one file. Using ArduinoJson v5 for now.
```

**Added**
```
Library building
Linting on pull request
```


tablatronix:
This combines both platforms?

shafr:
it has both esp8266 & esp32 library builds & example builds

tablatronix:
Thanks!

tablatronix:
merged and fixed up 
stupid question why are the boards reversed?

esp32:
        run: pio ci --board=nodemcuv2

shafr:
yep, thats bug, please reverse

tablatronix:
oddly it was working just fine

shafr:
also you can check my 'canary' repo actions - i have one that calculates FW / disk usage for pull requests

tablatronix:
interesting, I will check it out, I need to work on doxygen

tablatronix:
looks like we have a mix of yml and yaml, which one is best more standard?
I always use .yaml

shafr:
There's no rule, probably I was not paying attention, I prefer using shortest one.

PR #1319, opened 2021-12-14
switch handleWiFi to chunkedResponse

bwjohns4:
This is actually my first PR, hopefully I'm doing this correctly. I was using a lot of parameters and sometimes they were getting dumped out of the response in handleWiFi(). This is how I fixed the issue.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -1248,6 +1248,11 @@ void WiFiManager::handleWifi(boolean scan) {
     WiFi_scanNetworks(server->hasArg(F("refresh")),false); //wifiscan, force if arg refresh
     page += getScanItemOut();
   }
+
+  server->chunkedResponseModeStart(200, FPSTR(HTTP_HEAD_CT));
+  server->sendContent(page);
+  page = ""; //clear for next send
+
   String pitem = "";
 
   pitem = FPSTR(HTTP_FORM_START);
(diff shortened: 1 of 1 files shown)


bwjohns4:
I modified this to use the underlying functions within chunkedResponseModeStart() and chunkedResponseFinalize() which appear to also be in the ESP32 webserver. The ESP32 library also has the sendContent() functions, only in ESP8266 there is the extra start and finalize functions. I haven't tested this yet, but it should be a 1 for 1 swap as far as functionality goes.

tablatronix:
Thanks!, I started working on this, you beat me to it

tablatronix:
can you if you get a chance add
page += getParamOut(); back and see if this still works?

tablatronix:
I made a branch, using page refs and chunked is turned off, curious if it also works
https://github.com/tzapu/WiFiManager/tree/feature_pagestringsref

tablatronix:
@bwjohns4 have you had a chance to test this branch to see if this also works ( instead of chunked?)

tablatronix:
bump ,wondering if you tested this test branch to see if it helps instead

bwjohns4:
@tablatronix , this looks like it should work for the problem I experienced. I was just using lots of parameters with some of them being custom_html (which took a fair bit of memory just on its own) so I had to send them chunked. I was using around 46 parameters total (20 that were just HTML text chunks).

F1p:
I have the page hanging on send with chunked type

alexhopeoconnor:
This earlier chunked-response direction is relevant to the proposal in #1873. The maintained fork now uses async/chunked portal rendering to avoid constructing complete portal pages as large Strings, but it is a substantial refactor rather than a safe standalone patch. #1873 proposes treating the work as a new major version, restoring multi-language support, and defining a deliberate compatibility boundary.

Issue #1327, opened 2021-12-26
Adding the option to start/keep running the portals in the background after connecting


axlan:
This is a really simple feature request. I'd like to use this project both for it's Wifi management, but also as a general configuration page. To do this I run the class in non-blocking mode, and I want the config portal to keep running even after the WiFi connection completes successfully. I've made a simple PR to support this for my own use, but you may be interested in merging it upstream.

axlan:
I actually built this out a bit with a full example of how I'd be using it:
https://github.com/axlan/extended-wifimanager

tablatronix:
This feature is not needed due to non blocking mode already existing

axlan:
> This feature is not needed due to non blocking mode already existing, ondemand webportal existing, and this would only work on enableconfigportal and autoconnect failure. So I do not see the utility here, as opposed to manually starting configportal or webportal manually.
> 
> I have other requests for leaving cp open after save, which makes a bit more sense to me.
> 
> Maybe I am misunderstanding this entirely?

I initially hit this before I realized that the settings could be shown on their own page. In that usage mode saving the settings closes the web portal. It was also hard to enter settings in that mode since they would clear the WiFi credentials.

In non-blocking mode with the settings on their own page, I don't think this issue is relevant. I guess you can consider changing the behavior for the case with non-blocking mode, but the settings share the WiFi config page.

tablatronix:
Both of these should not happen.
- [ ] Empty creds should not wipe anything
- [ ] nonblocking should not stopconfigportal on saves

tablatronix:
do you happen to have logs of this ?

tablatronix:
I could not reproduce saves breaking saved creds

axlan:
I'm using the master commit: `810f144`

It looks like the main issue (clearing the credentials on an empty save) is fixed.

The configuration portal still closes on connection, which is what I originally proposed a setting to disable.

Here are the logs I captured in testing.

Here's an example main.cc I'm using for reference:

```cpp
#include <WiFiManager.h> // https://github.com/tzapu/WiFiManager
WiFiManager wm;
WiFiManagerParameter custom_field; // global param ( for non blocking w params )

void setup() {
    WiFi.mode(WIFI_STA); // explicitly set mode, esp defaults to STA+AP    
    // put your setup code here, to run once:
    Serial.begin(115200);
    
    //reset settings - wipe credentials for testing
    //wm.resetSettings();

    // test custom html(radio)
[...]

PR #1328, opened 2021-12-26
Add parameter to let the portals keep running in background.

axlan:
This is a really simple mod I made to support using this project both for it's Wifi management, and also as a general configuration page. To do this I run the class in non-blocking mode, and I want the config portal to keep running even after the WiFi connection completes successfully. I also made the corresponding issue: https://github.com/tzapu/WiFiManager/issues/1327 .


axlan:
I added some additional commits I needed to support my particular use case ( https://github.com/axlan/extended-wifimanager ) . You may not want the parts that change the portal shutdown behavior, but I found using a `std::unique_ptr` for the `WiFiManagerParameter` memory management, made the interface a lot less error prone (I was hitting bugs when my vector of parameters would do move constructors).

PR #1331, opened 2022-01-06
Enable Portal Timeout in not Blocking Portal

ricardollreis:
Notice that if we set the _setConfigPortalTimeout( **60** )_ and _setConfigPortalBlocking( **false** )_ we got the first message 
"*wm:[2] Portal Timeout In 60 seconds" but nothing after that. 

This PR is calling the handling for the timeout inside of the process function, in other to utilize the timeout mechanism 

Tested with ESP8266


tablatronix:
I think there was an open issue on this, but I cant find it, thanks!

PR #1344, opened 2022-01-24
Both infoids and _ap_password are intialized by check time.

pplanel:
Comparing it to NULL raises operation overload.


tablatronix:
error: 'class String' has no member named 'isEmpty'

tablatronix:
reverted

Issue #1341, opened 2022-01-22
WiFiManagerParameter does not handle move constructor


axlan:
#### Hardware
WiFimanager Branch/Release:  Master commit 810f144

Esp8266/Esp32: 

Hardware:			D1-mini lite

### Description

WiFiManagerParameter class does not properly handle the C++ move constructor resulting in a use after free error.

### Reproduction Example

```cpp
#include <vector>

#include <WiFiManager.h> // https://github.com/tzapu/WiFiManager

WiFiManager wm;
std::vector<WiFiManagerParameter> parameters;

void setup() {
    WiFi.mode(WIFI_STA); // explicitly set mode, esp defaults to STA+AP    
    // put your setup code here, to run once:
    Serial.begin(115200);

    // Allocate new space and construct parameter in it
    parameters.emplace_back("id1", "field1", "", 40);
    // Allocate new space, move "id1" and construct parameter after it
    parameters.emplace_back("id2", "field2", "", 40);

    // Add parameters to WiFiManager
    for (auto& param : parameters)
    {
      wm.addParameter(&param);
    }

    wm.setConfigPortalBlocking(false);
    //automatically connect using saved credentials if they exist
    //If connection fails it starts an access point with the specified name
    if(wm.autoConnect("AutoConnectAP")){
        Serial.println("connected...yeey :)");
    }
    else {
        Serial.println("Configportal running");
    }

    wm.startConfigPortal();
}

void loop() {
    wm.process();
}
```

I added a log message to record the alloc/free's:

```
new char allocated id1
new char allocated id2
char freed id1
*wm:[2] Added Parameter: id1
*wm:[2] Added Parameter: id2
```

### Solution

What's happening here is that the `std::vector` is calling the implicit move constructor: https://en.cppreference.com/w/cpp/language/move_constructor

This constructor doesn't correctly handle moving the `_value` data which is then freed. A similar issue is presumably why the copy constructor is made private.

There are three fixes that I see.

1. Delete the move constructor. My example doesn't build if I add: `WiFiManagerParameter(const WiFiManagerParameter&&) = delete;`
2. Implement a move (and optionally a copy) constructor. This is fairly straightforward and just needs to correctly propagate the pointer to the new instance and set the old instance to NULL so it isn't freed.
3. Switch to using smart pointers. I took this approach for a fork I made for testing. Here's the code:

```cpp

class WiFiManagerParameter {
  public:
    /** 
[...]

Issue #1050, opened 2020-05-11
Char* conversion for default_value errornous when adding WiFiManagerParameter();


zen85:
#### Hardware
**WiFimanager Branch/Release:**
- [ ] Master
- [X] Development (OTA)

**Esp8266/Esp32:**
- [X] ESP8266
- [ ] ESP32

**Hardware:			ESP-12e, esp01, esp25**
- [ ] ESP01
- [X] ESP12 E/F/S (nodemcu, wemos, feather)
- [ ] Other

**ESP Core Version:	2.4.0, staging**
- [ ] 2.3.0
- [ ] 2.4.0
- [X] 2.6.3
- [ ] staging (master/dev)

### Description

Hi, 

When we add a custom WiFiManagerParameter with a normal char* to initialize the default value it has a strange conversion error. Following code: 

```
char mqtt_broker[12] = "192.168.1.108";
custom_mqtt_broker = WiFiManagerParameter("mqttbroker", "MQTT Broker", mqtt_broker, sizeof(mqtt_broker));
Serial.println(custom_mqtt_broker.getValue());
Serial.println(mqtt_broker);`
```
puts out on the Serial: 
```
⸮⸮⸮?168.1.10
192.168.1.108
```

### Settings in IDE

Module:  			NodeMcu, Wemos D1

Additional libraries: 

### Sketch

```cpp

#include <FS.h>                   //this needs to be first, or it all crashes and burns...
#include <ArduinoJson.h>          //https://github.com/bblanchon/ArduinoJson
#include <WiFiManager.h> //https://github.com/tzapu/WiFiManager WiFi Configuration Magic
#include <PubSubClient.h>
#include <ESP8266WiFi.h>

#define USEOTA
// enable OTA
#ifdef USEOTA
#include <WiFiUdp.h>
#include <ArduinoOTA.h>
#endif

#define TRIGGER_PIN 0 //Pin used to start AccessPoint to configure device - Just ground GPIO for 2 seconds

WiFiManager wm;
ESP8266WebServer server(80);




WiFiClient espClient;
PubSubClient client(espClient);

long lastReconnectAttempt = 0;

boolean reconnect() {
  Serial.println("attempting connecting to mqtt");
  if (client.connect("arduinoClient")) {

    Serial.println("MQTT CONNECTED");
    // Once connected, publish an announcement...
    client.publish("outTopic", "hello world");
    // ... and resubscribe
    client.subscribe("inTopic");
  }
  return client.connected();
}

[...]

PR #1346, opened 2022-01-24
Move WiFiManagerParameter to use a smart pointer and add copy/move constructors

axlan:
This is to address https://github.com/tzapu/WiFiManager/issues/1341 and https://github.com/tzapu/WiFiManager/issues/1050

These changes should have no affect on the functionality except to slightly simplify the class, and add the additional copy/move functions.

Feel free to use this as a starting point if there are stylistic inconsistencies.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -24,17 +24,12 @@ uint8_t WiFiManager::_lastconxresulttmp = WL_IDLE_STATUS;
  * --------------------------------------------------------------------------------
 **/
 
-WiFiManagerParameter::WiFiManagerParameter() {
-  WiFiManagerParameter("");
+WiFiManagerParameter::WiFiManagerParameter(const char *custom) {
+  _customHTML     = custom;
 }
 
-WiFiManagerParameter::WiFiManagerParameter(const char *custom) {
-  _id             = NULL;
-  _label          = NULL;
-  _length         = 1;
-  _value          = NULL;
-  _labelPlacement = WFM_LABEL_BEFORE;
-  _customHTML     = custom;
+WiFiManagerParameter::WiFiManagerParameter(const WiFiManagerParameter& other) {
+  init(other._id, other._label, other.getValue(), other._length, other._customHTML, other._labelPlacement);
 }
 
 WiFiManagerParameter::WiFiManagerParameter(const char *id, const char *label) {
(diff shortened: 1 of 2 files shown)


tablatronix:
There was a failed check, I merged anyway into new branch

```
/Users/shawn/projects/microcontrollers/dev/libraries/WiFiManager/WiFiManager.cpp:83:13: error: 'make_unique' is not a member of 'std'
   _value  = std::make_unique<char[]>(_length + 1);
             ^
/Users/shawn/projects/microcontrollers/dev/libraries/WiFiManager/WiFiManager.cpp:83:30: error: expected primary-expression before 'char'
   _value  = std::make_unique<char[]>(_length + 1);
                              ^
```

tablatronix:
- [ ] fix make_unique for C++11

tablatronix:
Do you know a better way to rewrite this, or add an overload for make_unique, it looks pretty easy from what I read

axlan:
You can replace:
` _value  = std::make_unique<char[]>(_length + 1);`
with
 ` _value  = std::unique_ptr<char[]>(new char[_length + 1]);`

They're basically identical.

tablatronix:
So wait now that these are smartpointers we can probably better determine when they are out of scope right?

this is what i have in there now

```cpp

      Serial.println((String)_params[i]->_length);
      if (_params[i] == NULL || _params[i]->_length == 0 || _params[i]->_length > 99999) {
        // try to detect param scope issues, doesnt always catch but works ok
        #ifdef WM_DEBUG_LEVEL
        DEBUG_WM(DEBUG_ERROR,F("[ERROR] WiFiManagerParameter is out of scope"));
        #endif
        return "";
      }
    }
```

axlan:
If I understand what you're asking correctly, you want to detect when the external code lets its allocation for `params_` deallocate (go out of scope or explicitly delete).

The smart pointer I added her only manages  `WiFiManagerParameter::_value`. Basically, just automatically frees the memory whenever _value would be over ridden or the parent is deconstructed. It doesn't affect the management of pointers of `WiFiManagerParameter`.

If you want that to be managed with smart pointers you'd need the user to pass in a smart pointer (shared, or unique depending on how you want the ownership to work). I can make a PR that implements that as well, but at some point I guess it might be easier to copy the `WiFiManagerParameter` into an internal allocation instead of making the user allocate space for them. This is potentially more expensive (need to spend CPU cycles for the copy, and not let the user specify where the memory is allocated), but it would do a lot to simplify how these parameters are specified. I'll make a version of this later since I think it would be preferable to expecting the library user to directly deal with smart pointers.

tablatronix:
Well what alot of people do, is put the wm obj or params in setup, then if they try to do an ondemand in loop etc. It crashes.

Obviously this is because of how parameters are passed into the param class, and most folks just dont understand that much.or why their code crashes, I probably should see about adding compiler warnings also to reduce support issues

tablatronix:
Well I have on my roadmap to make the whole setup of params as you have, allow them to be containerized and looped, or even stored in a main struct that is mapped into fs back and fort easily.

PR #1355, opened 2022-02-11
added dynamic custom html and ExitConfigCallback

SolderingSoldier:


tablatronix:
What does this do ?

jonathanvanpeteghem:
@SolderingSoldier Thanks, I was looking for this ExitConfigCallback (and would have added it myself)

@tablatronix I would use it to restart the esp on exit. For the moment I have an assyncwebserver running. 
When I go into the config portal (via a push button) I first terminate the assync webserver. And then we go in the config portal.
But on return from the config portal,  I can only get the assync webserver working again by doing a reset. I am sure there are ways to "revive" the assync webserver, but I can't  see how. So the exit callback is useful to reset the esp, or maybe later to revive my "assync webserver".

Would you consider adding the exit callback by default?

I see you are working on an assync WM. I look forward to that.

tablatronix:
I think the main issue is that the port stays in use, even if we close the webserver, it seems to not free up properly as least that is what I found the last time I checked.

Yeah it works pretty well, I need to redo OTA and some stuff that works differently

I cannot merge this as it has other commits in this PR so I will close this and add it to the callback issue for consideration, thanks

PR #1369, opened 2022-03-06
Add softAP client connect/disconnect callbacks

davidmartinek-dev:
Add softAP client connect/disconnect callbacks. Can be useful when you want to do something when a client connects/disconnects to/from a softAP. For example, change led color, display a message on display, etc...


tablatronix:
can this be done with esp callbacks in nonblocking mode?

PR #1404, opened 2022-04-30
Silence debug output

mazgch:
This is probably a left over from debugging, can we remove or at least move it to a DEBUG_WM


tablatronix:
yup local testing, oops, thanks!

PR #1471, opened 2022-08-09
GH Actions Check

thorrak:
I noticed something when looking at GH actions that I think means the action doesn't accomplish what is expected.


thorrak:
Closing this PR, as it was intended to test github actions, not be merged.

PR #1475, opened 2022-08-10
Bring feature_asyncwebserver up to date with master

thorrak:
This PR brings feature_asyncwebserver up to date with master, and also gets the branch to a point where most of the tests compile.

Although there is a lot of work left to get this branch to the point where it could be incorporated back into master (setting async features behind feature flags, for example) this will hopefully make that process slightly easier.

### DIFF
--- a/.github/workflows/compile_examples.yaml
+++ b/.github/workflows/compile_examples.yaml
@@ -11,58 +11,58 @@ on:
       - '.github/workflows/compile_library.yml'
 
 jobs:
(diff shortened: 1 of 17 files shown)


thorrak:
Brought feature_async up to date again as of where master stood an hour ago. 

The example that is failing to compile is one that explicitly overrides `WebServer` - I haven't looked to see if there is a comparable override for Async in this case.

Issue #1502, opened 2022-09-21
Fail to build because for ESP8266 not have method getChipModel in ESP class.


thiti-y:
#### Hardware
WiFimanager Branch/Release:  Master

Esp8266/Esp32: ESP8266

Hardware:
ESP-12e

Core Version:
[v2.0.13-beta](https://github.com/tzapu/WiFiManager/releases/tag/v2.0.13-beta)

### Description

Fail to build because for ESP8266 not have method getChipModel in ESP class.

<img width="983" alt="Screen Shot 2565-09-21 at 2 11 32 PM" src="https://user-images.githubusercontent.com/7018802/191446130-a7fa12a0-b262-4be1-be61-395c8185c0d7.png">

### Settings in IDE

Module:
NodeMcu

Additional libraries:
N/A

### Sketch
N/A

### Debug Messages

<img width="983" alt="Screen Shot 2565-09-21 at 2 11 32 PM" src="https://user-images.githubusercontent.com/7018802/191446130-a7fa12a0-b262-4be1-be61-395c8185c0d7.png">

PR #1503, opened 2022-09-21
fix: not found getChipModel method for ESP8266

thiti-y:
Fail to build because for ESP8266 not have method getChipModel in ESP class.

Screen Shot 2565-09-21 at 2 11 32 PM

Resolve: #1502


tablatronix:
Oops

PR #1507, opened 2022-10-02
Update README.md

diraniyoussef:
1) typo 
2) clarify a feasible way to install this library since I couldn't find this library in Arduino IDE Library Manager


tablatronix:
This library is in the library manager

Issue #1109, opened 2020-08-16
[FR] Add support for Joining WPA(2) Enterprise Networks (cert upload etc)


aenertia:
#### Hardware
**WiFimanager Branch/Release:**
- [X] Development

**Esp8266/Esp32:**
- [X] ESP8266
- [X] ESP32

**Hardware:
- [X] Other

**ESP Core Version:
- [X] staging (master/dev)

### Description

WPA Enterprise modes are possible of both the esp32 and the esp8266 but require certs for the network to be embedded into the eprom to be secure (there are some caveats with how CHAP/PEAP user credentials are exchanged, these are documented in several threads on getting WPA enterprise to work)

This feature request is to allow WiFiManager the ability to upload a new cert and join a WPA Enterprise Network. This is especially useful for a huge amount of use cases, including IoT Vlan Dynamic Assignment, Campus Sensornets and many other situations. Making this functionality available via WiFiManager will take the complication of setting up the ESPs to use WPA Enterprise trivial.

Issue #404, opened 2017-08-30
WPA2 Enterprise fails


MarcFinns:
I brought  a device of mine at work and I realised that it does not work with the GUEST network that uses WPA2 Enterprise. It does connect but IP is 0.0.0.0.
Anyone interested in this feature? Is it a known issue and can it be solved?
thanks all.

tzapu:
Hi,

maybe this is of help
https://github.com/esp8266/Arduino/issues/1102 <https://github.com/esp8266/Arduino/issues/1102>

cheers
alex

MarcFinns:
Hi, does it mean that it is now working..?

tablatronix:
#309

Issue #309, opened 2017-02-07
ass support for IEEE 802.1x


CWempe:
It would be great if WiFiManager would support authentication via [IEEE 802.1x](https://en.wikipedia.org/wiki/IEEE_802.1X).

In some universities there is no "normal" WiFi and you need to connect via 802.1x.
This would make developing in education much easier.

I understand this is supported with the new SDK 2.0.0 for esp8266.
Which is available for Arduino, I think.
https://github.com/esp8266/Arduino/commit/ae13809c8184300aab9e3f09ef23af23d936b7ee

tablatronix:
no where near stable though stable is still SDK 1.5.3
no milestone either, i have not even touched master in months, how stable is it ?

CWempe:
I have no idea if the new sdk is stable or not.
I just read somewhere that 802.1x is (or will be) supported with sdk 2.0.0.

l00mi:
Being new to this party I am wondering if the SDK 2.0.0 has matured and if 802.1x could be considered?

bkrajendra:
Im testing WPA2 PEAP with following code. will let you know if it works:


```
#include <ESP8266WiFi.h>

extern "C" {
#include "user_interface.h"
#include "wpa2_enterprise.h"
}


// SSID to connect to
static const char* ssid = "IOCARE_NEW";
// Username for authentification
static const char* username = "myraddisuser";
// Password for authentication
static const char* password = "rapass123456";

void setup(){
 // WPA2 Connection starts here
  // Setting ESP into STATION mode only (no AP mode or dual mode)
    wifi_set_opmode(STATION_MODE);
    struct station_config wifi_config;
    memset(&wifi_config, 0, sizeof(wifi_config));
    strcpy((char*)wifi_config.ssid, ssid);
    wifi_station_set_config(&wifi_config);
    wifi_station_clear_cert_key();
    wifi_station_clear_enterprise_ca_cert();
    wifi_station_set_wpa2_enterprise_auth(1);
    wifi_station_set_enterprise_identity((uint8*)username, strlen(username));
    wifi_station_set_enterprise_username((uint8*)username, strlen(username));
    wifi_station_set_enterprise_password((uint8*)password, strlen(password));
    wifi_station_connect();
  // WPA2 Connection ends here
}
```

bkrajendra:
Tried all day long... but no success. getting error as follows while connecting:


```
Welcome to WPA2 Eneterprise Test
15:45:26.038 -> 1
[...]

PR #1512, opened 2022-10-16
WPA2 Enterprise PEAP support

jameshi16:
Hi,

I'm currently working in an environment where I have to connect the ESP32 to a WPA2 Enterprise network protected with PEAP. Given the nature of such a network, the credentials are used for more than just WiFi. I was hence uncomfortable with configuring my username & password directly on a sketch that I would be sharing, so I decided to use WiFiManager.

The PR contains some code to get WiFiManager working with PEAP-protected networks. Is implementing support for WPA2 Enterprise networks currently on the roadmap?

I had to use my own NVS storage, as it seems like the ESP32 framework doesn't automatically save credentials for anything outside of a typical SSID & Password combo.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -1958,7 +1958,6 @@ void WiFiManager::handleWifiSave() {
                   #ifdef WM_DEBUG_LEVEL
                   DEBUG_WM(DEBUG_DEV,F("cannot write to nvs flash"));
                   #endif
-                      Serial.printf("bruh");
                }
 
               nvs_set_str(handle, "wm.pass", _pass.c_str());
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -16,6 +16,8 @@
 
 #ifdef ESP32
 uint8_t WiFiManager::_lastconxresulttmp = WL_IDLE_STATUS;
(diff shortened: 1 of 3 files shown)


jameshi16:
Hi @tablatronix, apologies for pinging on a closed PR, I just wanted some feedback for learning purposes :sweat_smile: 

Was the implementation too naive? In hindsight, perhaps it was better to submit a PR that supports all the different WPA2 Enterprise configurations.

I see that there is some interest in WPA2 Enterprise in #309, #404, and a completed feature request at #1109. Is implementing WPA2 Enterprise of any form within the roadmap? Will it be incremental (support one authentication method at a time), or would it be a large feature dump?

Thanks for looking into this, really appreciate what you have done with WiFiManager.

tablatronix:
I have enterprise wifi lab/sandbox now, I will see about working on this. I would maintain this PR as a fork maybe if anyone else wants it working out of the box, but I am hesitant to hardcode any memory storage outside of native esp methods.

Maybe if we can find a way in the future in to pass in a struct or pointer to fs and be able to auto handle params, but this can also be done by extending the param class.

tablatronix:
Created branch for this in case anyone wants to use it

PR #1555, opened 2023-01-27
Only reconnect if WiFi updated

presslab-us:
If only the parameters have been updated then don't try to reconnect on save. Otherwise `_ssid` and `_pass` are blank (because "placeholder" input attribute is used) and the connection will fail.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -1858,6 +1858,7 @@ void WiFiManager::handleWifiSave() {
   else {
     page = getHTTPHead(FPSTR(S_titlewifisaved)); // @token titlewifisaved
     page += FPSTR(HTTP_SAVED);
+    connect = true; //signal ready to connect/reset process in processConfigPortal
   }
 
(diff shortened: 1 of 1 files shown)


presslab-us:
still a problem in the latest code

tablatronix:
What browser? placeholders should not be submitted in inputs

presslab-us:
Yes, exactly, the placeholder is not submitted, and therefore WiFiManager uses the returned (blank) SSID and password to reconnect.  This causes the connection to fail.

My patch makes it so that if the SSID and password are not changed it will not try to reconnect with blank `_ssid` and `_pass`.

Tested with Chromium 110.0.5481.177

tablatronix:
If its blank then it shouldnt be connecting at all

presslab-us:
Is that the intended behavior?  Every time that you go into the settings page you need to type in the ssid and password again, even if the placeholder is correct?

tablatronix:
No, I will take a look, something else seems to be broken also

PR #1557, opened 2023-01-30
added web flasher to repo 

mihalixu:
#1339 good first Issue


tablatronix:
What else is needed? Do we need actions to build binaries?

PR #1581, opened 2023-04-06
Added method to set custom sub title

Fruchtzwerg94:
- Allows to set a custom subtitle like done with the title as well
- Uses AP name and IP as fallback like done before
- Fixes _@todo use hostname?_


Fruchtzwerg94:
Good hint @ab-tools ,
I've merged the latest changes of the master to the branch and added the subtitle to the proposed sites as well.
BR

PR #1595, opened 2023-04-21
Make *Show Password* a label so one can click on it

Wandmalfarbe:
It is now possible to toggle the checkbox by clicking on the label *Show Password*.


graue70:
Hi,
I noticed you only changed the template but the code in `wm_strings_en.h` and `wm_strings_es.h` don't contain this change. How is your change supposed to get into those files?

tablatronix:
Ill fix it

PR #1596, opened 2023-04-22
Allow removing custom parmeters from config/web portal 

PotatoCider:
This allows the ability to remove custom parameters from the config/web portal. The user may want to remove this for security reasons (i.e. only allow custom parameters to be set when in web portal mode)


tablatronix:
I am going to add some security to the menu options for this I think , so then ota and params etc can only be enabled on web side.. Or we can add a callback and the decision can be made in user code at runtime.. not sure yet

PR #1637, opened 2023-07-13
Limited redirect

raomin:
When I connect to wifimanager from my windows, it generates many requests (eg wpad/wpad.dat, autodiscover, etc.).
All these requests end up being served by wifimanager twice: once for redirect and a second time as 404.
It's often so consuming for asyncweb that the esp32 ends up crashing and my pc disconnected.
It's often better from my mobile where I get less extra requests.

However, I believe we shouldn't redirect these extra requests in the first place. With this change, I see no more crashes nor disconnections.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -2420,6 +2420,15 @@ boolean WiFiManager::captivePortal() {
   bool doredirect = serverLoc != server->hostHeader(); // redirect if hostheader not server ip, prevent redirect loops
   // doredirect = !isIp(server->hostHeader()) // old check
   
+  //Only redirect the portal detection requests. Avoids overload.
+  doredirect = doredirect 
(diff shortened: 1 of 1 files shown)


raomin:
Note: I took the list of redirection detection from [here](https://github.com/RavenSystem/esp-homekit-devices/blob/8693ec8bf673a2ea5a6a9d618118c858e8cb74b2/sdk/esp-open-rtos-rsf/extras/libesphttpd/libesphttpd/util/captdns.c#L221). I tested with windows, ios and android.

tablatronix:
Yeah you are supposed to turn off all data when using captive portals but adding a filter is a good idea problem is keeping up with it there are no standards and they all randomly change

tablatronix:
Add custom or callout for captive checking

PR #1657, opened 2023-09-20
Update WiFiManager.cpp

furynick:
Fix some syntax errors that prevent compiling within Visual Studio


olonsoft:
The if statement needs `==` not `=`

Issue #1654, opened 2023-09-04
Checkbox example?


Dario-Ciceri:
Hi, could you please provide a working checkbox example?
I tried many examples, included yours but checkbox value is always ""

Thank you.

woodenplastic:
i set up a eventlistener in wm_stings_en.h in the script part to set the value of the checkbox to true or false depending if its checked. 

"<script>";
............
"function setCheckboxEventListener(input) {"
"  input.addEventListener('change', function() {"
"    this.value = this.checked ? 'true' : 'false';"
"  });"
"}"
"document.addEventListener('DOMContentLoaded', function() {"
"var input = document.getElementById(HERE YOUR INPUT ID OF THE CHECKBOX);"
"input.value = input.checked ? 'true' : 'false';"
"setCheckboxEventListener(input);"
"});"
"</script>";

Issue #1639, opened 2023-07-14
add setCustomHtml(const char* custom)  -> WifiManagerParameter


iPhilBln:
Is there a way to mark checked boxes? 
For example I have some const parameters which can only be clicked. 

```cpp
const char* checkbox_unchecked = "type=\"checkbox\"";
const char* checkbox_checked   = "type=\"checkbox\" checked";

const char* tu_berlin_ip    = "130.149.7.7";

tu_berlin_ntp_server WiFiManagerParameter;
new (&tu_berlin_ntp_server) WiFiManagerParameter("tu_berlin_ntp_server", "TU Berlin", tu_berlin_ip, strlen(tu_berlin_ip), checkbox_unchecked, WFM_LABEL_AFTER);
```
At the momemt I see no chanced to see if the marker was set or not. Correct me if I am wrong but it would be nice to edit the custom html part in the callBack function. But there is also a issue with this values, because if the user doesn't click on an value it will be deleted. 
I've discribed it here: #1111

Laxilef:
+

Issue #1111, opened 2020-08-20
Value for Checkboxes overwritten - "Advanced" example


Lithimlin:
#### Hardware
**WiFimanager Branch/Release:**
- [ ] Master
- [x] Development

**Esp8266/Esp32:**
- [ ] ESP8266
- [x] ESP32

**Hardware:			ESP-12e, esp01, esp25**
- [ ] ESP01
- [ ] ESP12 E/F/S (nodemcu, wemos, feather)
- [x] Other: TTGO T8

**ESP Core Version:	2.4.0, staging**
- [ ] 2.3.0
- [x] 2.4.0
- [ ] staging (master/dev)

### Description

I've tried using the checkboxes (as in the "Advanced" example) to enable and disable things on my ESP. While doing that I figured out that once you *disable* a checkbox and exit the configuration or let the AP timeout, the value saved in the `WiFiManagerParameter` is lost and the checkbox will not work again.

#### How to reproduce

1. Open the "Advanced" example 
2. Enable the checkbox (line 30) and commented the radioboxes (lines 33f)
3. Upload the sketch
4. Connect to the AP
5. In the `Params`, click `save`*without* enabling the checkbox
6. Close the AP or let it timeout
7. Open the AP again by pressing the button
8. Try enabling the checkbox and saving

The value of the checkbox will remain empty even if it is checked

### Sketch
"Advanced" example:

```cpp

/**
 * WiFiManager advanced demo, contains advanced configurartion options
 * Implements TRIGGEN_PIN button press, press for ondemand configportal, hold for 3 seconds for reset settings.
 */
#include <WiFiManager.h> // https://github.com/tzapu/WiFiManager

#define TRIGGER_PIN 0

WiFiManager wm; // global wm instance
WiFiManagerParameter custom_field; // global param ( for non blocking w params )

void setup() {
  WiFi.mode(WIFI_STA); // explicitly set mode, esp defaults to STA+AP  
  Serial.begin(115200);
  Serial.setDebugOutput(true);  
  delay(3000);
  Serial.println("\n Starting");

  pinMode(TRIGGER_PIN, INPUT);
  
  // wm.resetSettings(); // wipe settings

  // add a custom input field
  int customFieldLength = 40;


  // new (&custom_field) WiFiManagerParameter("customfieldid", "Custom Field Label", "Custom Field Value", customFieldLength,"placeholder=\"Custom Field Placeholder\"");
  
  // test custom html input type(checkbox)
[...]

PR #1662, opened 2023-10-09
Add virtual for getValue (for override getValue)

Laxilef:
Hi
Currently there is no way to create a dynamic checkbox. I wrote the code for this:
```C++
class CheckboxParameter : public WiFiManagerParameter {
public:
  const char* checked = "type=\"checkbox\" checked";
  const char* noChecked = "type=\"checkbox\"";
  const char* trueVal = "T";

  CheckboxParameter(const char* id, const char* label, bool value): WiFiManagerParameter("") {
    init(id, label, value ? trueVal : "0", 1, "", WFM_LABEL_BEFORE);
  }

  const char* getValue() const override {
    return trueVal;
  }

  const char* getCustomHTML() const override {
    return strcmp(WiFiManagerParameter::getValue(), trueVal) == 0 ? checked : noChecked;
  }

  bool getCheckboxValue() {
    return strcmp(WiFiManagerParameter::getValue(), trueVal) == 0 ? true : false;
  }
};
```

But for it to work, getValue must be overridden.

https://github.com/tzapu/WiFiManager/issues/1654
https://github.com/tzapu/WiFiManager/issues/1639
https://github.com/tzapu/WiFiManager/issues/1111

### DIFF
--- a/WiFiManager.h
+++ b/WiFiManager.h
@@ -211,7 +211,7 @@ class WiFiManagerParameter {
     // WiFiManagerParameter& operator=(const WiFiManagerParameter& rhs);
 
     const char *getID() const;
-    const char *getValue() const;
+    virtual const char *getValue() const;
     const char *getLabel() const;
     const char *getPlaceholder() const; // @deprecated, use getLabel
     int         getValueLength() const;


sbquinlan:
I'm lost. Can you explain the example a bit? You're setting the default to "T" and making getValue always return "T", but then only calling base getValue(). How does this create a dynamic checkbox?

Laxilef:
For a checkbox the value must always be filled (1/yes/true/T). State of the checkbox is the attribute "checked". Therefore, we override the getValue method that is used to generate the form:
https://github.com/tzapu/WiFiManager/blob/f328cd05cd5aa758b414d8e6961e21a1f6370cab/WiFiManager.cpp#L1777C9-L1777C81

And we get the real value using getCheckboxValue. Then you can change the value and the checkbox in the form will be checked or unchecked.

tablatronix:
I thought this was already addressed. Thanks will take a look

Laxilef:
> I thought this was already addressed. Thanks will take a look

What do you think about it? :)
So far I have to use a fork in my open source projects.

Issue #1044, opened 2020-05-02
https portal


Nimo11:
Hello,

I will find it very useful to be able to use the portal in https. I use the non-blocking portal and it would be more secure. Do you have it in your road map.

Thank you for this very useful work

tablatronix:
Yes ideally you would provide your own websever and pas it into wm

Nimo11:
You could maybe change handleRequest to this :

```
void WiFiManager::handleRequest() {
  if (WiFiManager::AuthenticationRequired&&!server->authenticate(WiFiManager::AuthUser,WiFiManager::AuthPassword))
  {
    server->requestAuthentication(HTTPAuthMethod::DIGEST_AUTH);
  }
  _webPortalAccessed = millis();
}
```

tablatronix:
What’s that do?

Nimo11:
It add authentication ability. 

Changing handleRequest with this code and adding in header file this property :

  bool AuthenticationRequired;
  const char* AuthUser;
  const char* AuthPassword;

So developers can decide to ask authentication simply.

Sorry for this too fast reply.

tablatronix:
Does this work with captive portal ?

Nimo11:
Yes, all requests are processed. But the behavior can be modified by changing the value of AuthenticationRequired after the autoconnect

tablatronix:
Ok this definitely does not work for captive portals. I am gonna try to add a landing page and see if that works.

Nimo11:
I try it this week and it work me with dev branch. I restart with acces point and i have login needed, After configuration i have a login page on my portal. What's wrong with your test?

tablatronix:
Https authenticaiton does not work on my devices. It never pops up

Issue #982, opened 2019-12-12
Secure web routes endpoints


tablatronix:
Add security to endpoints, make sure things are not available, menu can be hidden but we do not disable the handles, users can still use all forms etc.

PR #1669, opened 2023-10-23
Add HTTP Digest Authentication

PotatoCider:
This PR adds HTTP Digest Authentication (or revives the dead feature). May not work with some Android native webview browsers (Upon tapping "Sign in to network" notifications).

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -1296,30 +1296,25 @@ void WiFiManager::HTTPSend(const String &content){
 }
 
 /** 
- * HTTPD handler for page requests
+ * handles pre-request authentication
+ * return false if not authenticated
  */
-void WiFiManager::handleRequest() {
+bool WiFiManager::handleRequest() {
   _webPortalAccessed = millis();
 
-  // TESTING HTTPD AUTH RFC 2617
-  // BASIC_AUTH will hold onto creds, hard to "logout", but convienent
-  // DIGEST_AUTH will require new auth often, and nonce is random
-  // bool authenticate(const char * username, const char * password);
-  // bool authenticateDigest(const String& username, const String& H1);
-  // void requestAuthentication(HTTPAuthMethod mode = BASIC_AUTH, const char* realm = NULL, const String& authFailMsg = String("") );
-
-  // 2.3 NO AUTH available
-  bool testauth = false;
-  if(!testauth) return;
+  // We use DIGEST_AUTH since connection is over insecure HTTP
   
+  if (!_enableAuth) return true;
+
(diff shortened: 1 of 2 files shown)


Laxilef:
This is what is sorely missing right now.
I hope this PR will be merged.

tablatronix:
Yes! How much testing have you done?

PotatoCider:
I'm using this feature currently in a smart door lock, it works well for my use case. So far I've not encountered any bugs other than the captive portal not supporting HTTP auth prompts so a regular web browser would be required to login.

tablatronix:
- [ ] Breaks captive portal detection OSX, how can we add options here
- [ ] Add ability for auth to select pages, OTA, ERASE

Laxilef:
> Add ability for auth to select pages, OTA, ERASE

I don't think this is necessary. This will be confusing.
After all, we can dynamically turn authentication on and off in code.

> Breaks captive portal detections, how can we add options here, can we omit root perhaps, or add an auth  splash page

I see in the library code that the captivePortal() method is called before the handleRequest() method, so this should not affect the operation of the captive portal.

Also, I think captive portal should be disabled for WebPortal because it breaks port forwarding. In my projects I solved it like this:
```c++
      if (wm.getWebPortalActive()) {
        wm.setCaptivePortalEnable(false);
      }
```

tablatronix:
That is already another issue, #982

Laxilef:
> Good point about cp on webportal, surpised this hasnt been noticed before, how does it break port forwarding, got a test case?

You can flash any sketch and try to forward port 8181 -> 80 through the router. This can also be easily checked using the wokwi emulator.
I noticed this by chance when one of the users encountered a problem and I asked him to provide access to the ESP.
When I opened IP:8181, a redirect (302) to IP:80 occurred.
https://github.com/tzapu/WiFiManager/blob/18145e42ba686f297d4e741fb8805dffcb99a920/WiFiManager.cpp#L2451

Laxilef:
> I gave up on this previously #1044, as it also did not work with cp at all on OSX/IOS devices, there is no auth in the cp browser

But how? Authorization is requested after processing the CP:
```c++
  if (captivePortal()) return; // If captive portal redirect instead of displaying the page
  if (!handleRequest()) return;
```

I'll check this in a couple of days.

tablatronix:
CaptivePortal() is just a redirect, so it goes to handleroot / which is auth and an invalid response for the cp browser.

I was looking at the port code yeah we are not checking host port from header, that might be fixable, but makes sense to just to skip cp checks entirely in not on ap.. since its irrelevant

tablatronix:
I never addressed this because its dumb, it would be worth implementing basic html auth instead to protect in ap/cp mode.

PR #1670, opened 2023-10-25
Update wm_strings_en.h

DanielLester83:
moddified const char HTTP_STYLE[] 
added h1 and h3 to line 91
created a new .footer class on new line 96
example use setCustomHeadElement("<div class='footer'>Some Text</div>");


tablatronix:
Can you add this to the template html ?

DanielLester83:
> Can you add this to the template html ?

[Like This](https://github.com/DanielLester83/WiFiManager/blob/85e96eaf39160c3ae5b8f08818568f7e9338b1d4/extras/WiFiManager.template.html)

Is that what you wanted?

tablatronix:
Can you add that change to the PR?

PR #1673, opened 2023-11-05
cutting down

happysploots:


tablatronix:
I have this, I forgot I never pushed it, So i reworked it, let me know if you have anything to add

https://github.com/tzapu/WiFiManager/tree/feature_slimable

PR #1695, opened 2024-01-01
Custom HTML at the Top of "<body>"

ab-tools:
Added setCustomTopBodyElement allowing to define custom HTML to be added at the top of the "<body>" tag.

This is very useful to add a custom HTML header, e. g. also a logo at the top of the page.

I know this would in general be possible with `WIFI_MANAGER_OVERRIDE_STRINGS` and copying the whole `wm_strings_en.h` file, however, if the only goal is to add a single line of HTML code at the top of the page this seems like a bad idea:
Whenever something would get added in that file, these changes would need to be copied over manually to the custom file again.

Having the option to just "inject" and additional HTML code at the top of the body tag (similar as it is already possible to inject something at the bottom of the "head" tag) seems a much better option.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -1288,6 +1288,7 @@ String WiFiManager::getHTTPHead(String title){
     page += FPSTR(HTTP_HEAD_END);
   } 
 
+  page += _customTopBodyElement;
   return page;
 }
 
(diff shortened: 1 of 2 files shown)


anishsane:
Since you have already done this, may I suggest a change to add a footer as well? (setCustomFooterElement)

I am currently using a hack to add a footer using javascript like this:

```  wm.setCustomHeadElement("<script>setTimeout(function(){document.querySelector('.wrap').appendChild(document.createTextNode('Build time: " __DATE__ "/" __TIME__ "'));}, 100);</script>");
```
(I could not use the timestamp from the info page, since it is a timestamp of when the wifi-manager library files were compiled. For repeated builds, old library .o files are reused.)

But having a separate setCustomFooterElement function would be a much simpler approach.

dmadison:
This is great! I hope you don't mind, I reworked the name as "body header" and added a corresponding "footer" function, PR #1769.

dmadison:
This should now be closed (superseded by #1769). Thank you!

PR #1698, opened 2024-01-07
Allow subclassing

sinedied:
Allows to extend WiFiManger and parameter classes so you can customize it.

Making the internal variables/methods protected instead of private should not change any existing behavior.


tablatronix:
Thanks, any ideas or thoughts for an example to include?

sinedied:
Sure, my own use case could be a good example: reuse internal methods to build your own HTTP server, using either STA or AP mode (reusing mode of the startConfigPortal() method)

tablatronix:
- [ ] add short example sketch

PR #1720, opened 2024-03-13
Adde methods to se if static IP addresses were configured, and what those addresses are

jonasbjurel:
To be able to find out if the network with static addresses or with DHCP following methods are proposed:

    //check if static network configuration is set
    bool          isStaticIp(void);

    //provides static host ip address provided method isStaticIp() returns true
    void          getStaticIp(IPAddress* staticIp);

    //provides static gateway ip address provided method isStaticIp() returns true
    void          getStaticGw(IPAddress* staticGw);

    //provides static network mask provided method isStaticIp() returns true
    void          getStaticSn(IPAddress* staticSn);

    //provides static DNS ip address provided method isStaticIp() returns true
    void          getStaticDns(IPAddress* staticDns);

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -2784,6 +2784,61 @@ void WiFiManager::setSTAStaticIPConfig(IPAddress ip, IPAddress gw, IPAddress sn,
 }
 
 /**
+ * [isStaticIp description]
+ * @access public
+ * @param {[type]} void
+ * @return bool true if static network configuration was configured, false if DHCP was configured
+ * provides static or dhcp network configuration
+ */
+bool WiFiManager::isStaticIp(void) {
+    return _sta_static_ip ? true : false;
+}
+
+/**
+ * [getStaticIp description]
+ * @access public
+ * @param {[type]} IPAddress* staticIp
+ * @return void
+ * provides static host ip address provided method isStaticIp() returns true
+ */
+void WiFiManager::getStaticIp(IPAddress* staticIp) {
+    *staticIp = _sta_static_ip;
+}
+
+/**
+ * [getStaticGw description]
+ * @access public
+ * @param {[type]} IPAddress* staticGw
+ * @return void
+ * provides static gateway ip address provided method isStaticIp() returns true
+ */
+void WiFiManager::getStaticGw(IPAddress* staticGw) {
+    *staticGw = _sta_static_gw;
+}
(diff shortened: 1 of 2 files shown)


namix71:
Do you have any plans to update it to a product that supports static IP settings?

tablatronix:
This is all stuff you can do in code already

jonasbjurel:
> This is all stuff you can do in code already

Could you just give a hint on how?
For instance understand if the addresses where provided as static addresses in the wifi-manager, or if it was provided by the DHCP server?

jonasbjurel:
A small codesnip:
`    if (networkConfig.staticIp = wifiManager.isStaticIp()) {
        wifiManager.getStaticIp(&networkConfig.ipAddr);
        wifiManager.getStaticSn(&networkConfig.ipMask);
        wifiManager.getStaticGw(&networkConfig.gatewayIpAddr);
        wifiManager.getStaticDns(&networkConfig.dnsIpAddr);
    }
    else {
        networkConfig.ipAddr = WiFi.localIP();
        networkConfig.ipMask = WiFi.subnetMask();
        networkConfig.gatewayIpAddr = WiFi.gatewayIP();
        networkConfig.dnsIpAddr = WiFi.dnsIP();
    }`

tablatronix:
Static configs are not stored in flash, so user code has to store and recall them. So no need for wm specific wrappers

jonasbjurel:
> Static configs are not stored in flash, so user code has to store and recall them. So no need for wm specific wrappers

That is exactly why I want to know where to fetch the config, I.e. using: bool isStaticIp(void);
I guess my small additions does not break anything, and I got my code working with these additions, which I failed to without them.
BR/J

tablatronix:
I am assuming this only works right after you save wifi ?

jonasbjurel:
> I am assuming this only works right after you save wifi ?

Correct, after having saved WiFi, I use these method to check if a static address was configured, and if so I fetch the static addresses using these methods, if not static I fetch the DHCP assigned addresses using the normal WiFi methods.
BR/Jonas

jonasbjurel:
Hi again, what is the status, will my proposed changes be merged?
I'm approaching a release of one of my projects which rely on WiFi manager, if these changes are not merged I'm afraid I need to maintain a fork of the WiFi manager.
Regards /Jonas

PR #1727, opened 2024-03-29
Renaming as Arduino needs same name for sketch and directory.

iharshadec:
Example under NonBlocking --> OnDemandNonBlocking was giving error as .ino name was onDemandNonBlocking.
Corrected .ino name with Capital O as there should be same .ino name as directory.


iharshadec:
@tablatronix  this was very simple just making sure file name is same. Just renamed file.

PR #1750, opened 2024-07-14
Removal of conditionals before every debug output

khReichel:
Removal of redundant compiler conditionals before each debug output.

The conditionals remained at code positions that had further functionality in addition to the debug output


tablatronix:
Why?

khReichel:
Because there has been a ticket for that taks, right?

tablatronix:
I tried testing this in the min size branch, and I could not confirm that the compiler was not still allocating flash strings.

khReichel:
hmm.. I do not find any requirement about allocating flash strings in the ticket?
So maybe we have to extend the ticket about that topic?

PR #1755, opened 2024-08-06
FR Translation

AByGCreation:
Proposal of French translation

### DIFF
--- a/wm_consts_fr.h
+++ b/wm_consts_fr.h
@@ -1,0 +1,266 @@
+/**
+ * wm_consts.h
+ * internal const strings/tokens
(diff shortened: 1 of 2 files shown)


AByGCreation:
I propose a French translation for the portal.

Waz-Cpt:
I think there is an error with the change.
Line 3368 is currently
_debugPort.printf("[MEM] free: %5lu | max: %5u | frag: %3u%% \n", free, max, frag);

I believe it should be
_debugPort.printf("[MEM] free: %5u | max: %5u | frag: %3u%% \n", free, max, frag);

tablatronix:
Yeah there was a change there, I would have to go back and check

PR #1757, opened 2024-08-13
Italian translation

GMagician:
I didn't added a wm_consts_it.h (I saw it only in Frech translation but I suspect it's not required)


tablatronix:
where is the pr?

tablatronix:
I guess I can merge it manually

Issue #1660, opened 2023-09-29
ESP8266 boot loop


tablatronix:
Exception Cause: 3  [LoadStoreError: Processor internal physical address or data error during load or store]

void WiFiManager::WiFiManagerInit(){
  // setMenu(_menuIdsDefault);

causes exception

tablatronix:
#1587 moved menutokens in progmem

So this probably has to be adjusted
` if((String)menuitem == String(_menutokens[j])){`

tablatronix:
tentative fix, not sure what the exact issue or fix here should be

f328cd05cd5aa758b414d8e6961e21a1f6370cab

PR #1763, opened 2024-09-04
Solving problem of access violation.

tarcisiojr:
In certain scenarios, the ESP8266 was unexpectedly triggering an access violation error. After conducting a thorough debugging process, I identified and resolved the issue with this code update. The solution is based on guidance from the official ESP8266 documentation, specifically regarding the use of `PROGMEM` for handling program memory. You can find more details in the documentation here: [ESP8266 PROGMEM Documentation](https://arduino-esp8266.readthedocs.io/en/latest/PROGMEM.html#:~:text=32%20bit%20aligned.-,Functions%20to%20read%20back%20from%20PROGMEM,-%C2%B6).

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -1438,9 +1438,9 @@ void WiFiManager::handleParam(){
 String WiFiManager::getMenuOut(){
   String page;  
 
-  for(auto menuId :_menuIds ){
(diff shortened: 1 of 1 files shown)


tablatronix:
Thanks, I forgot all about this bug. #1660

PR #1764, opened 2024-09-16
Add Simplified Chinese Translation

wzsx150:
1. Add Simplified Chinese Translation: zh-CN.
2. Fix some translation issues.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -2057,7 +2057,7 @@ void WiFiManager::handleInfo() {
   }
   page += F("</dl>");
 
-  page += F("<h3>About</h3><hr><dl>");
+  page += getInfoData("abouthead");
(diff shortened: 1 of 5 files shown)


wzsx150:
Usage:
You need to add the statement #define WM_STRINGS_FILE "wm_strings_cn.h" in the build system. Note that it should not be added directly in the .ino file.
I added this statement in the board definition file, for example, in the Linux version of Arduino at the path /home/user/.arduino15/packages/esp32/hardware/esp32/3.0.4/variants/esp32s3/pins_arduino.h

Does anyone have a better way to enable the Chinese translation file?

wzsx150:
Can someone merge this PR?

tablatronix:
Will not merge, contains non lang file changes

PR #1772, opened 2024-10-28
WiFiManagerParameter max length and encapsulation

dmadison:
This PR performs a few fixes around the `WiFiManagerParameter` class to improve clarity, usability, and encapsulation:

* The `length` parameter has been changed to `maxLength` for all class functions
    * The parameter refers to the max length of the form and the size of the internal buffer, not the length of the string passed as an argument. The internal name (`_length`) has *not* been changed to maintain compatibility with user code, as it's exposed as a protected value and through friendship with `WiFiManager`.
* `setValue()` now has a version without a max length argument
    * The original 'length' argument was confusing as a user, since it seemed to refer to the length of the string passed in while it's actually the max length of the buffer / value in the form. This new function allows for changing the value without changing the max length.
* `getValueLength()` has been changed to `getValueMaxLength()`
    * This better represents what is being returned, since it's the maximum length of the form / buffer and not the length of the value (i.e. what's *currently* in the buffer, and what's returned from the webpage). The original function is marked deprecated and has not been changed for compatibility reasons, but it would be nice if it returned the length of the value in the next major release.
* Friendship between  `WiFiManagerParameter` and `WiFiManager` has been removed
    * The few internal calls in `WiFiManager` have been changed to use the public functions and there is no remaining reason for the friendship to exist. Removing the friendship boosts encapsulation.
* `setValue()` now checks for a negative max length
    * All of the corresponding 'length' arguments should probably be unsigned (`size_t`). Again, this was not changed so-as not to break backwards-compatibility.

This PR should have **no** breaking changes. The few items marked deprecated should be removed on the next major release.

---

If this is merged, here are the changes to make on the next major release:
* `WiFiManagerParameter::_length` should be refactored to `_maxLength`
* `WiFiManagerParameter::getValueLength() const` should be removed
* All `WiFiManagerParameter` `maxLength` arguments should use `size_t` instead of `int`

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -86,6 +86,11 @@ void WiFiManagerParameter::setValue(const char *defaultValue, int maxLength) {
     // Serial.println("cannot set value of this parameter");
     return;
   }
+
+  if(maxLength < 0){
+      // Serial.println("cannot set length below zero");
+      return;
+  }
   
   // if(strlen(defaultValue) > maxLength){
   //   // Serial.println("defaultValue length mismatch");
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -76,6 +76,10 @@ WiFiManagerParameter::~WiFiManagerParameter() {
 //   return *this;
 // }
 
+void WiFiManagerParameter::setValue(const char *defaultValue) {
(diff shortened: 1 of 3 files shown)


dmadison:
Closing this in anticipation of further changes.

PR #1774, opened 2024-10-30
WiFiManagerParameter Encapsulated HTML Generation

dmadison:
This PR encapsulates the `WiFiManagerParameter` HTML generation as part of the class, rather than as a function within `WiFiManager`. This allows derived classes to define their own HTML output and greatly increases encapsulation:

```cpp
 virtual String getHTML() const;
```

This also greatly simplifies the `getParamOut()` function of `WiFiManager`:
```cpp
...
    // add the extra parameters to the form
      page += _params[i]->getHTML();
    }
...
```

This is a **non-breaking** change and should not modify the output HTML, with the exception of `{I}` as noted below.

---

Note that this PR contains a potentially breaking change, by removing token I (`T_I` / `{I}` / capital 'i') processing for parameters which include that token in their string. This was removed because the parameter HTML is no longer generated in the context of the loop, and I didn't see a compelling reason to pass the iterator to the class function.

From what I can see parsing through the history of the `strings_en` file that feature was never used by the library itself, but it could potentially affect a user somewhere with custom string files (`HTTP_FORM_LABEL` / `HTTP_FORM_PARAM`). If that's objectionable I'm happy to revert that commit (c4665fc6cca7344795e0a88fc9ca4eccaa84085e).

Please note that this PR incorporates the necessary API changes from #1773.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -1748,8 +1748,8 @@ String WiFiManager::getParamOut(){
     char valLength[5];
 
     for (int i = 0; i < _paramsCount; i++) {
-      //Serial.println((String)_params[i]->_length);
-      if (_params[i] == NULL || _params[i]->_length > 99999) {
+      //Serial.println((String)_params[i]->getValueMaxLength());
+      if (_params[i] == NULL || _params[i]->getValueMaxLength() > 99999) {
         // try to detect param scope issues, doesnt always catch but works ok
         #ifdef WM_DEBUG_LEVEL
         DEBUG_WM(WM_DEBUG_ERROR,F("[ERROR] WiFiManagerParameter is out of scope"));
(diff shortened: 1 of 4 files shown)


tablatronix:
I think I started adding a param index, so parameters didn't need an explicit id/name in the case of dynamic or stateful params etc

tablatronix:
Also why did you change id() to name(), wont this break anyones implementation of a param child class ?

dmadison:
>I think I started adding a param index, so parameters didn't need an explicit id/name in the case of dynamic or stateful params etc

That makes sense. Although with this change to move HTML generation to the parameter class, I'd suggest that the decision for how to handle a lack of a name/id should be the responsibility of whatever derived class chooses to omit it.

>Also why did you change id() to name(), wont this break anyones implementation of a param child class ?

No, the change from 'id' to 'name' should not break API compatibility.

The API changes were submitted separately in #1773. Those commits were included here because this also modifies the parameter class, and not including them would otherwise create a conflict when (if) both PRs are merged.

tablatronix:
Yeah I recall now, my bad, finally reviewing this

PR #1818, opened 2025-05-02
Fix watchdog timeout on esp32 when config portal is active

h2zero:
Calling `yield()` on the esp32 doesn't let the idle task run and therefore doesn't feed the watchdog. Instead, calling `delay(1)` forces the context change and allows the idle task to run.

### DIFF
--- a/WiFiManager.cpp
+++ b/WiFiManager.cpp
@@ -812,7 +812,7 @@ boolean  WiFiManager::startConfigPortal(char const *apName, char const *apPasswo
 
     if(!configPortalActive) break;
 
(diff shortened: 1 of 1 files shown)


thorrak:
Huh, I never knew this was a thing: https://esp32.com/viewtopic.php?t=809#p3442

I always assumed that yield() fed the watchdog/allowed _all_ pending tasks to run, not just higher priority ones. Go figure.

h2zero:
Yep, just like yielding when driving, don't need to stop if nobody else has the right of way.

Funny that this hasn't been merged or commented on at least until now.

PR #1849, opened 2026-02-22
Improve SSID selection interface

caiohamamura:
* Moved <a> to wrap the entire row including signal icons so the user can click anywhere on the div to select the SSID.


cniedzi:
@tablatronix   After the last commit, once I've select WiFi SSID, the password text box gets disabled and it is not possible to enter WiFi password.

[https://github.com/tzapu/WiFiManager/issues/1850](https://github.com/tzapu/WiFiManager/issues/1850)

PR #1865, opened 2026-05-16
fix: ESP32-S3/C3 softAP clients unable to connect

### LINKED ISSUE #1482: Esp32 s3/c3 softap not working 2.3.0
There are several open issues that all seem related to wifi being broken on esp32 s3 and c3, seems to work on the latest esp lib. Not sure if its a combination on wifimanager starting and stopping sta mode the ln starting softap making this hard to debug.

schulzh:
I'll be honest upfront: I've let opencode create this PR, since I do not have knowledge how the wifi stack works. However, the result works for me on an ESP32-S3, so I want to share this anyhow.

This was the result of multiple rounds of patching (or rather have the AI patch) and testing the connection manually with my mobile phone, so it seems to me that all three issues were valid and needed fixing, but more testing might be needed.

---

On ESP32-S3 and ESP32-C3, the config portal AP starts but clients cannot complete WiFi association. This is caused by:

1. Rapid WiFi mode switching (OFF->STA->AP) leaving the AP interface unstable on these chips.
2. The preloaded async WiFi scan causing channel hopping that disrupts AP client connections.
3. The STA disconnect/disable sequence before startAP() interfering with the subsequent AP initialization.

Fixes:
- Use clean mode transitions with delays in startConfigPortal() for ESP32-S3/C3 (WIFI_OFF -> 500ms -> WIFI_AP -> 500ms -> softAP)
- Use WiFi channel 6 instead of 1 (channel 1 is unstable under DNS+HTTP traffic load on ESP32-S3)
- Skip preloaded WiFi scan on ESP32-S3/C3 (scan on-demand instead)
- Skip STA disconnect sequence on ESP32-S3/C3 before AP start

Tested on ESP32-S3-DevKitC-1 with espressif32 @ 6.12.0. Resolves #1482


roberttco:
Just wanted to confirm that this PR fixes the erradic behavior I was having using an ESP32-C3-SuperMini

LotusOttawa:
> Just wanted to confirm that this PR fixes the erradic behavior I was having using an ESP32-C3-SuperMini

Not quite. I had to turn down the TX power on some ESP32-C3 SuperMini boards.

